PR #4, opened 2018-09-06
Use long description as Markdown

auscompgeek:
This avoids a dependency on pypandoc when building the Python package,
and fixes the long description not rendering correctly on PyPI.

See <https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi>.


benfred:
thanks for this! sorry for the long delay in merging -

PR #20, opened 2018-09-07
Fixes #7 Update error msg for missing dependency

### LINKED ISSUE #7: Additional dependency on perl
I found a small hiccup:
When using py-spy with the -flame option, it also depends on Perl. Not having any Perl installation, trying to execute -flame netted me a "Could not execute Perl. File not found [oserror 2]" error. After installing Perl it functioned flawlessly.

Would be cool if this dependency could be mentioned in the readme to make troubleshooting easier.

kamyarrasta:


hatgit:
is there any common solution for resolving this error when running ```sudo py-spy --``` from MacOS terminal: 
"No such file or directory (os error 2)"

kamyarrasta:
@hatgit I don't know what is the relation between the error that you mentioned here and this pull request which is about missing Perl dependency.
Not sure what does this command mean: `sudo py-spy --` 
As far as I know the usage is `py-spy --duration <duration> --pid <pid> --rate <rate>`

benfred:
@hatgit I think this is because when you run py-spy as sudo, the sudo user can't find the py-spy binary. you can either install py-spy to root (sudo pip install py-spy), or you can preserve the path by going ```sudo env PATH=$PATH" py-spy -- python progam.py'```

@kamyarrasta thanks for the PR! I'm leaving this open for the time being, because this code is going to go away soon. I'm working on creating a common crate for py-spy/rbspy (https://github.com/rbspy/rbspy/issues/180) and this file is going to be part of it. Once the new crate is created I will apply this patch there

benfred:
With the latest version we don't require perl anymore - closing

PR #52, opened 2018-10-04
Add PYSPY_CROSS_COMPILE_TARGET parameter to setup.py.

frederikaalund:
This small change allows the users to cross compile py-spy for a given target system.

I've used this patch to compile py-spy for armv7 via buildroot.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,12 +54,16 @@ class PostInstallCommand(install):
         # issues like https://github.com/benfred/py-spy/issues/5.
         # Note: we're only doing this on demand since this requires musl-tools installed
         # but the released wheels should have this option set
+        cross_compile_target = os.getenv("PYSPY_CROSS_COMPILE_TARGET")
         if os.getenv("PYSPY_MUSL_64"):
             compile_args = " --target=x86_64-unknown-linux-musl"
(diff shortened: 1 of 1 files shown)


frederikaalund:
> Unfortunately it doesn't look like we can distribute binary wheels for arm targets: https://stackoverflow.com/a/47018230 yet. I'm not shipping a sdist package, so the best way to distribute this for arm is probably using cargo =(

Source distribution is fine for me for less common targets such as ARM. [Buildroot](https://buildroot.org) (which I'm using) prefers to build everything from source anyhow. [Yocto](https://www.yoctoproject.org/) would probably prefer a pre-built binary though.

> Have you tested this pacakge against arm? I don't have a raspberry-pi or other arm processor to test out so I'm interested in hearing if you got this going =)

Yes! Py-spy compiled without errors and worked on the first run on my [STEMlab 125-14](https://www.redpitaya.com/f130/STEMlab-board). It's a Zynq 7010 with an ARM Cortex A9 (ARMv7-A architecture). Comparable to an iPhone 4 in terms of processing power.
Py-spy seems as fast and responsive on this board as it does on the desktop. I do, however, get this message in the top of py-spy:

    Duration { secs: 5, nanos: 69075235 } behind in sampling, results may be inaccurate. Try reducing the sampling rate.

PR #61, opened 2018-11-16
Stop to convert readme to rst format

4383:
Now pypi display properly the markdown format so doesn't need
to continue to convert markdown to rst


4383:
Duplicated with https://github.com/benfred/py-spy/pull/4

PR #91, opened 2019-02-16
First draft embedded web server

benfred:
This adds a basic webserver to serve up realtime profile results. These
results can be visualized using d3 / d3-flamegraph, and all the assets
are compiled into the binary using rust-embed and then served up with
rouille.


boxed:
What's the status of this? It sounds awesome!

benfred:
@boxed There is a dev release that you can try out here: ```pip install py-spy==0.4.0.dev0```.  Once you have it installed you can launch by going ```py-spy serve -- python myprogram.py``` to launch the webserver.

You should be able to get a flamegraph while the program is running - and select different time ranges to show in the flamegraph. You can also drill down into a specific function, and see the timings inline with the source code for that function - but that feature needs a ton of work right now.

PR #121, opened 2019-06-24
Use Cargo.toml values for clap app parameters

toabctl:
Instead of duplicating the information like version, appname and
description, just use the values provided by the Cargo.toml file.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,8 +9,6 @@ serialize =
 
 [bumpversion:file:setup.py]
 
-[bumpversion:file:src/config.rs]
-
 [bumpversion:file:Cargo.toml]
 search = name = "py-spy"
 	version = "{current_version}"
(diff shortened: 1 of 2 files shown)


toabctl:
@cc benfred

toabctl:
cc @benfred

benfred:
Thanks for the PR! 

I'm using bump2version to increment the version strings, so this line https://github.com/benfred/py-spy/blob/master/setup.cfg#L12 also needs to be removed.

I'm also considering relicensing this project from GPL3 to MIT (I want to share a bunch of code here w/ rbspy - and need a compatible license for that, also a MIT license will make adoption at my work a little easier). Are you good with releasing your change under a MIT license?

toabctl:
> I'm using bump2version to increment the version strings, so this line https://github.com/benfred/py-spy/blob/master/setup.cfg#L12 also needs to be removed.

Done

> I'm also considering relicensing this project from GPL3 to MIT (I want to share a bunch of code here w/ rbspy - and need a compatible license for that, also a MIT license will make adoption at my work a little easier). Are you good with releasing your change under a MIT license?

Yes. Feel free to relicense to MIT.

Issue #112, opened 2019-05-10
FreeBSD support


akhramov:
Hi, thanks for your awesome application, I really love it.

FreeBSD is a somewhat popular system among developers and it would be nice to have it supported.

Do you have any objections so far? If not, keep reading.

As far as I can tell, this application relies on few third-party libs which do not support FreeBSD yet.
So, if we were making a checklist it would be like this:

- [ ]   Support FreeBSD in  [rbspy/proc-maps](https://github.com/rbspy/proc-maps).
**Status:** PR was created (rbspy/proc-maps#3).
- [ ]   Support FreeBSD in  [luser/read-process-memory](https://github.com/luser/read-process-memory).
**Status:** Support was added in luser/read-process-memory#7, however, I'd wait for luser/read-process-memory#10 for better ergonomics.
- [ ] Support FreeBSD here, in py-spy

Am I missing something? What do you think?

benfred:
Awesome! This would be great to add.

I think you've done most of the hard work here already (proc-maps/read-process-memory changes). Most of the remaining changes here should be in the 'remoteprocess' subcrate here.

Minimally we're looking at adding code to:
- [ ] lock the remote process (probably can use ptrace like in linux?)
- [ ] get the executable name of the process

Also - it would be great if we could also add code to do this (but isn't necessary to get basic python profiling working on freebsd):
- [ ] list threads of the remote process
- [ ] get if a thread is running or not
- [ ] lock an individual thread

I haven't ever used freebsd - but I think the tricky thing here is that there isn't procfs that we can use reliably (please correct me if I'm wrong =).

akhramov:
Thanks, Ben.

It shouldn't be a lot of hassle to implement all of this. In fact, I think that `ptrace(2)`  would handle most of the cases.

Alright, I'm working on the `remoteprocess` subcrate. Will get back with the PR. :smiley:

benfred:
Awesome! 
 
I thought about this a little more today - and I think there is some more work that needs to happen though. We'll need to get CI setup so that I don't accidentally break this in a future change, also we need to disable the native extension code otherwise this won't build easily.

I've been planning on adding a feature flag for the native extension code (also needed to re-enable the ARM build again https://github.com/benfred/py-spy/issues/89), so you won't need to implement that code for freebsd. I will get this going this weekend and update here once it's done.

[...]

PR #131, opened 2019-07-10
[#112] FreeBSD support

akhramov:
FreeBSD is a somewhat popular system among developers and it would be
nice to have it supported.
In order to support it, we have to patch the remoteprocess subcrate a
bit, as well as bump `proc-maps` and `read-process-memory`
crates version.

This change

[remoteprocess crate]

Adds FreeBSD support for

* process and thread locking (via ptrace)
* getting thread registers (via ptrace)
* getting process' executable and cwd (via libprocstat)
* read_process_memory contract has changed in the latest version, so
  OSX / Linux / Windows code has been adjusted accordingly.

The only thing that is left to be implemented is unwinder /
symbolificator. Therefore `py-spy` on FreeBSD does not support native
stack traces, yet.

[py-spy]

* Adds FreeBSD-specific versions of platform-dependent functions.
* Adds Cirrus CI configuration to run tests against FreeBSD
  platform. In order to set CI up, please refer to
  https://cirrus-ci.org/guide/quick-start/


akhramov:
@benfred 
Regarding dependencies... 

I have changed `read-proc-memory` to point to my fork.  The fork incorporates your changes from luser/read-process-memory#6 and my from luser/read-process-memory#10. I would be happy to have a better solution.

Also, `proc-maps` currently resolves to a git branch containing changes from rbspy/proc-maps#3. I believe, you can review / merge this one.

akhramov:
Alright, windows tests seem to fail. I'm looking into this.

akhramov:
The tests are green!

remoteprocess/Cargo.toml
```diff
@@ -2,19 +2,13 @@
 name = "remoteprocess"
 version = "0.1.0"
 authors = ["Ben Frederickson <github@benfrederickson.com>"]
-repository = "https://github.com/benfred/py-spy"
-homepage = "https://github.com/benfred/py-spy/tree/master/remoteprocess"
-description = "cross platform api for getting information on a running processes"
-readme = "README.md"
-license = "GPL-3.0"
-build="build.rs"
```
benfred: This block shouldn't be removed

remoteprocess/src/freebsd/lock.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use super::Error;
+use super::ptrace;
+
+#[derive(Debug)]
+pub struct ProcessLock {
+    pid: pid_t,
+}
+
+impl ProcessLock {
```
benfred: With linux - we're locking each thread individually, using multiple ptrace attach calls: https://github.com/benfred/py-spy/blob/1717b91d7fafc6910e1e9b76e3397a7d58e883dc/remoteprocess/src/linux/mod.rs#L61-L75

Does attaching to only the main thread pause all the child threads as well?

If not, can we share the linux ptrace locks w/ bsd here?

benfred:
Thanks for this! Mostly looks good. Also left a minor comment on the proc-maps PR.

One other thing - I'm considering relicensing this project from GPL3 to MIT (I want to share a bunch of code here w/ rbspy - and need a compatible license for that, also a MIT license will make adoption at my work a little easier). Are you good with releasing your change under a MIT license?

remoteprocess/src/freebsd/lock.rs
akhramov: > Does attaching to only the main thread pause all the child threads as well?

Yes, in FreeBSD, `PT_ATTACH` sends `SIGSTOP` to the whole process:
https://github.com/freebsd/freebsd/blob/7f19ffe6f947e40299a4be93b7cf320da2e17951/sys/kern/sys_process.c#L955


I test this behavior indirectly here:
https://github.com/benfred/py-spy/blob/d79a69a4f6f380a179944106126ce88265c4bab1/remoteprocess/src/freebsd/mod.rs#L254-L261

In  contrary, Linux `ptrace` operates solely on threads:

>  Therefore,  "tracee"  always  means  "(one)  thread", never "a (possibly multithreaded) process".  Ptrace commands are always sent to a specific tracee using  a  call  of  the form

So, BSD ptrace gives us more flexibility here.

remoteprocess/Cargo.toml
akhramov: Sure, my bad!

akhramov:
Thanks for the review!

I'm fine with releasing my code under MIT license. Moreover, I'm in no rush, you can accept this contribution after the relicensing.

akhramov:
Bumped proc-maps version. Also, I created a PR to your `read-process-memory` fork (benfred/read-process-memory#2).

I want you to keep control over `py-spy` dependencies. :)

benfred:
thanks!

drozdowsky:
thank you

PR #172, opened 2019-09-30
Fix integration_test on Windows

domartynov:
On my Windows 10 `python3` points to Windows Store app, not Python:

```
C:\>where python3
C:\Users\domar\AppData\Local\Microsoft\WindowsApps\python3.exe

C:\>where python
C:\Users\domar\AppData\Local\Programs\Python\Python37\python.exe
C:\Users\domar\AppData\Local\Microsoft\WindowsApps\python.exe
```

### DIFF
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -9,7 +9,7 @@ struct TestRunner {
 
 impl TestRunner {
     fn new(config: Config, filename: &str) -> TestRunner {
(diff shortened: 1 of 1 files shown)


benfred:
Thanks for the PR! I refactored things a bit differently though - I wanted to let people use the integration tests to also test out python2 if that is what's on their path : https://github.com/benfred/py-spy/pull/177 . This means we're using 'python' for all OS now - since this PR isn't necesesary

PR #174, opened 2019-10-03
Make python_binary optional to be None if no .data section like in .NET host

### LINKED ISSUE #171: "Failed to find .data section in PE binary" error when .exe hosts Python without .data
It looks like a .NET host `.exe` is missing the `.data` section. Can we make [`python_binary`](https://github.com/benfred/py-spy/blob/1842a68ac588ae8e44af226d2451c3cb65a902a8/src/python_spy.rs#L620) `Option<BinaryInfo>`, it seems `libpython_binary` is used for `python.exe` on Windows anyway (at least when I debug Python 3.7). Then, to replace [`expect`](https://github.com/benfred/py-spy/blob/1842a68ac588ae8e44af226d2451c3cb65a902a8/src/binary_parser.rs#L118) with `Err` leading to `None` in `python_binary`. Not sure how difficult to propagate the change, if ok, I can try to come up with a PR.

domartynov:
Fixes #171 by making `python_binary` optional (`None` on `Err` returned by `parse_binary`) and propagating the change. 

I have a change to test on Windows only though.

### DIFF
--- a/src/binary_parser.rs
+++ b/src/binary_parser.rs
@@ -112,15 +112,16 @@ pub fn parse_binary(filename: &str, addr: u64, size: u64) -> Result<BinaryInfo, 
                 }
             }
 
-            let data_section = pe.sections
+            pe.sections
                 .iter()
                 .find(|ref section| section.name.starts_with(b".data"))
-                .expect("Failed to find .data section in PE binary");
+                .ok_or_else(|| format_err!("Failed to find .data section in PE binary of {}", filename))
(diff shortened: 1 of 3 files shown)


domartynov:
Sample .NET host project I used to test against:
https://github.com/domartynov/pynethost, and compiled binaries:
https://github.com/domartynov/pynethost/files/3684309/pynethost.zip

benfred:
thanks!

Issue #124, opened 2019-07-07
Profile subprocesses


benfred:
It would be nice to be able to profile all the sub-processes of a python process. This would let us profile programs that use multiprocessing or gunicorn worker pools.

This will also help with profiling virtualenvs on windows w/ python 3.7.2 (https://github.com/benfred/py-spy/issues/81#issuecomment-477445626)

ygormutti:
I was thinking about a universal approach for profiling single and multiprocess applications, based on https://github.com/nylas/nylas-perftools approach. The idea is:

1. Somehow make py-spy hook into subprocesses. Maybe monkey patching `fork`? I think explicitly instrumenting application code would be acceptable if it's simple enough, like a one-liner at the app entrypoint, but I think there may be a better, less intrusive way;
2. Add an output option to export raw profiling data into some dir, the same data that would be used to build a flamegraph;
3. Create a `f"{pid}.json" file for each process into that dir and dump current data to them once in a while;
4. Add a subcommand to aggregate multiple raw data files into a single flamegraph;

I think this is the simplest approach to solve it in a way that would be useful to profile the projects I'm working on in a production environment. What do you think? I'd love to contribute with this.

benfred:
I think there are a couple of ways we can get the child processes here - but it's pretty OS depedendant:

1) Linux

We could try to use PTRACE_O_TRACEFORK to get notified when the process forks. The author of pyflame tried this route with https://github.com/uber/pyflame/issues/67 and sounds like he got pretty far with this approach, but hit some limitations of sigtimedwait that blocked him.

Alternatively we could periodically poll for subprocesses using procfs . The only annoyance here is that I think we can only get the *parent* of a process easily - so we'd have to scan each /proc/PID/stat file to get the parent and then filter out to ones that have a parent/grandparent etc that is the target program. This is almost certainly the easiest thing to get going, but isn't necessarily the most efficient method. This method is also what [rbspy does to profile subprocesses](https://github.com/rbspy/rbspy/blob/c8d5ac48a27618d7fbeb59434b137b6e8ef3788f/src/ui/descendents.rs#L84-L102)

2. Windows

[...]

PR #185, opened 2019-10-19
#124 [WIP] Profile subprocesses

akhramov:
It would be nice to be able to profile all the sub-processes of a
python process. This would let us profile programs that use
multiprocessing or gunicorn worker pools.

This change

* [py-spy] Adopts rbspy-like model of monitoring child processing via
  threads, communicating via mpsc channels.
* [remoteprocess] Borrows rbspy's process child detection
  functionality.

### DIFF
--- a/remoteprocess/src/linux/children.rs
+++ b/remoteprocess/src/linux/children.rs
@@ -1,0 +1,121 @@
+// This code is adapted from rbspy:
+// https://github.com/rbspy/rbspy/blob/3d09e3ced011eb10ab7a0f5906659820043e177e/src/ui/descendents.rs
+// licensed under the MIT License:
+/*
+MIT License
+
+Copyright (c) 2016 Julia Evans, Kamal Marhubi
+Portions (continuous integration setup) Copyright (c) 2016 Jorge Aparicio
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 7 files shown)


akhramov:
Alright, this is my POC take on #124. With this patch, one can profile python sub-processes on Linux.

However, I have stumbled upon a few obstacles, listed bellow.

**Implementation**
* [Top view] Since we receive samples from different threads, `py-spy top` view may update too often. Should we throttle it's update rate somehow?

* [Record] It is impossible to predict how many samples we are going to do across all processes. Thus we cannot set max value for the progress bar. Should we update the max value dynamically based on current number of known processes?

**Misc / tests**
- How do I test `main.rs` code with integration tests?


@benfred may I ask you to provide me some guidance?

benfred:
Thanks for the PR!

I’ve also been working on adding this support, I pushed out my changes here https://github.com/benfred/py-spy/pull/186

I started out sampling like you are here, and hit those issues you mentioned. I fixed by only generating 1 sample event for all processes, instead of having each python process generate samples independently from each other.  The idea here is that we still have 1 thread to profile each python process, but we collect all the output from these threads and collate it down so that we are still sending a predictable number of samples to the top/record viewers.  This fixed the problems with top/record you saw, and also let us figure out the correct %-active values in the top view.

I also wanted to show the complete process hierarchy in the flame graphs, so the output from the remoteprocess call is a little different and includes the parent pid for each child pid. The top view also shows the number of active processes

Anyways, I have it implemented for linux/osx right now in #186
I also tweaked the freebsd process code so that it can moved between threads there.

For testing, we don’t have anything that hits the code in main.rs yet =(. I’ve been moving code to other modules to test it out (like the multiprocess code is mostly in src/sampler.rs right now).

akhramov:
> The idea here is that we still have 1 thread to profile each python process, but we collect all the output from these threads and collate it down so that we are still sending a predictable number of samples to the top/record viewers

I went the similar route for the top view:

https://github.com/benfred/py-spy/pull/185/files#diff-639fbc4ef05b315af92b4d836c31b023R114-R128

I pushed the latest code I have, idk if it's going to be useful :) In fact, we made almost identical changes to `src/stack_trace.rs`

Since you are working on sub-processes profiling support, I am going to close PR. 

Is there something I can help you with? Are you planning to roll up support for all platforms at once?

benfred:
I think I got the changes for this almost done - but I do have some questions about the support you added for windows to rbspy that I was wondering if you could help me out with.

I couldn't get NtGetNextProcess to work like you did in rbspy - it seems to return every process created after the process, rather than just child processes of the process. Did you hit this problem? Or am I doing something wrong here: https://github.com/benfred/py-spy/pull/186/commits/b202391f73a96e1a39f1645fcfa1f74d7ef50faa#diff-ec6cd7ecae4554ff492bd3b29175e362R142-R166 . I'm a little worried that this problem might also exist in rbspy =(

I fixed by getting the parenet pid for each process returned, and then filtering down to ones that are descended from the target pid using NtQueryInformationProcess

akhramov:
Yeah, I’m  afraid you’re right.  I’m going to submit the  fix  to rbspy, thanks!


I also wonder  how big the performance impact of the official process retrieval  way (
https://docs.microsoft.com/en-us/windows/win32/toolhelp/taking-a-snapshot-and-viewing-processes)

PR #191, opened 2019-11-09
Include symbols from dynsym table of ELF bin/lib

jarus:
Some Linux distributions e.g. openSUSE deliver the python binary and
libpython with an empty ELF symtab. With this change py-spy also uses
the dynsym table for locating the relevant Python symbols.


benfred:
thanks!

PR #192, opened 2019-11-11
Remoteprocess: rbspy amendments.

akhramov:
Currently, rbspy implements a function which copies vector of arbitrary
structures of type U from remote process. Therefore, if we want rbspy
to leverage remoteprocess crate, we would need remoteprocess to
implement such function.

This change

* Implements `ProcessMemory#copy_vec` method, which copies a
  series of bytes from another process into a vector of structures of
  type T.
* Disables unwinding on non x86_64 platforms altogether, since build
  fails on 32-bit windows

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -2,14 +2,14 @@ use std::env;
 
 fn main() {
     // copied from remoteprocess/build.rs because I couldn't find a way to share this
+    // We only support native unwinding on x86_64 platforms
+    if env::var("CARGO_CFG_TARGET_ARCH").unwrap() != "x86_64" {
+        return;
+    }
+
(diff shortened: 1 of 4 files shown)


akhramov:
Please see rbspy/rbspy/pull/238 for more context

benfred:
Thanks! 

I'm going to split out remoteprocess to it's own repo, and make the unwind code optional via a feature flag (so py-spy can include it, but rbspy can exclude it).  The unwind code causes some extra build difficulties on linux, and since rbspy doesn't need it right now I want to be able to disable it.  

Once I have that done I'll bump the version for remoteprocess and update crates.io

PR #208, opened 2019-12-18
Fix typo in handling of lnotab.

### LINKED ISSUE #190: Line numbers wrong with Python 3.8
After upgrading my local project to Python 3.8, the line numbers that py-spy reports are about 3-8 times too high (difference varies). This affects both the flamegraph and the speedscope output.

**Environment:**
Python 3.8.0
Windws 10 Pro, Version 10.0.18362 Build 18362
py-spy==0.3.0 (but I believe it affects earlier versions, too)

Note: This situation occurs when the program that is analyzed is running on Python 3.8 and it's independent of the Python environment that I start py-spy from. To clarify:

* Profiled code runs in: 3.7, py-spy runs in: 3.7 -> line numbers OK
* Profiled code runs in: 3.7, py-spy runs in: 3.8 -> line numbers OK
* Profiled code runs in: 3.8, py-spy runs in: 3.7 -> line numbers *wrong*
* Profiled code runs in: 3.8, py-spy runs in: 3.8 -> line numbers *wrong*

anntzer:
Closes #190, AFAICT, by fixing a simple typo in #196.
(Perhaps it would be easier to just read increment as i8 and cast it to i32?  Dunno what's idiomatic rust, but that would at least look more idiomatic in C to me.)

Repro:
```
import time

def f():
    [time.sleep(.1)  # Must be split over two lines to see the error.
     for _ in range(10)]

f()
```

### DIFF
--- a/src/stack_trace.rs
+++ b/src/stack_trace.rs
@@ -152,7 +152,7 @@ fn get_line_number<C: CodeObject, P: ProcessMemory>(code: &C, lasti: i32, proces
         // Handle negative line increments in the line number table - as shown here:
         // https://github.com/python/cpython/blob/143a97f6/Objects/lnotab_notes.txt#L48-L49
         if increment >= 0x80 {
-            increment -= -0x100;
+            increment -= 0x100;
         }
         line_number += increment;
         i += 2;


benfred:
Thanks for the fix! It's an embarrassing mistake I made there.

For casting,  I'm not sure what the most idiomatic way to do this in rust. We could use [std::mem::transmute ](https://doc.rust-lang.org/std/mem/fn.transmute.html) I guess - but the current implementation at least matches the lnotab implementation in python, and doesn't require unsafe code so I'm inclined to leave it for now

I'm going to add the example you've given to the test suite - thanks for including that here. This made it really easy to verify that the fix here is working and solves the problem in #190

anntzer:
I don't think we need to overly worry about the most idiomatic way to upcast an int :)

anntzer:
Hi, any chance for a new release?  Thanks!
Happy new year :)

benfred:
fix is in v0.3.2

anntzer:
thanks!

PR #215, opened 2020-01-17
Link to deprecation of macOS built-in Python

adamchainz:
Another argument to stop using `/usr/bin/python` that I think is useful to point out to users.


benfred:
Thanks!

PR #222, opened 2020-02-09
In record, don't capture output from child procs

benfred:
When launching a child process we were capturing stdout/stderr and not showing.
This makes sense for the 'top' view, but isn't the best default for 'record'.
Change to show stderr/stdout from the child process. Passing a '--capture'
flag restores the previous behaviour.


pquentin:
Awesome, thanks for this @benfred!

PR #237, opened 2020-05-09
Include python thread names in recording

jarus:
In comparison to single py-spy dump, a recording doesn't include the custom defined python thread name. While analysing the performance of an application with various threads, I spent the most time on identifying the thread so that I could correlate the recording with other observability data e.g. logs and distributed traces.

This change extends py-spy's recording capabilities by including the python thread name in each stacktrace struct. The speedscope output formatter uses the collected data to create sample profiles with the thread name as profile name so that the thread name is visible in the speedscope UI.

### DIFF
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -1,5 +1,5 @@
 extern crate py_spy;
-
+use std::collections::HashSet;
 use py_spy::{Config, PythonSpy, Pid};
 
 struct ScriptRunner {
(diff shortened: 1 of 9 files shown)


tests/integration_test.rs
```diff
@@ -117,6 +117,34 @@ fn test_long_sleep() {
...
+
+    let traces = runner.spy.get_stack_traces().unwrap();
+    assert_eq!(traces.len(), 11);
+
+    let mut expected_threads: HashSet<String> =
+        (0..10).map(|n| format!("CustomThreadName-{}", n)).collect();
+    expected_threads.insert("MainThread".to_string());
+    let detected_threads: HashSet<String> = traces
+        .iter()
+        .map(|trace| trace.thread_name.as_ref().unwrap().clone())
```
benfred, line 143: The unwrap on this line is causing issues on travis with the x86_64 linux variants:  https://travis-ci.org/github/benfred/py-spy/builds/684995442 .  I believe this is probably because we are using a python2 interpreter there - and we can only get thread names from version 3.6 on.

Can you only tun this on python 3.6+ ? There is code in test_local_vars that does this

jarus:
> One potential problem with this is that the thread names can change on the python thread, and we'd still use the old value since we're caching the thread names.

Yes, I also thought about this limitation but I actually never saw a Python application that changes the thread name after the initial start.

What do you think, is it acceptable to include the thread name determination overhead by default? The sampling of applications with frequent changing threads could be negatively impacted.

benfred:
Yeah - I think I'd rather not incur the performance penalty. Thanks again!

PR #283, opened 2020-08-12
fix typos

basnijholt:


README.md
```diff
@@ -89,16 +89,15 @@ associated with each stack frame by setting the ```--locals``` flag.
 
 ### Why do we need another Python profiler?
 
-This project aims to let you profile and debug any running Python program, even if the program is
+This project aims to let your profile and debug any running Python program, even if the program is
```
kylepollina, line 92: `profile` here is used as a verb, and not referencing to _your_ profile. It lets you <verb>profile</verb> your code. So this actually isn't a correct revision

benfred:
Thanks for the PR! I appreciate pointing out some of the mistakes here -

There was a couple things I disagree with (like behaviour being misspelt  and 'your profile' vs 'you profile') - so I didn't merge directly.  I fixed the appropriate typo and the some of the other capitalization issues in the most recent commit when I was changing the readme anyways. 

thanks again!

Issue #229, opened 2020-03-08
Make '--output' optional


Nudin:
When I use py-spy I often forget the output parameter. I also often don't care much about the filename and often it would be nice to know the date of the profile.
We could solve all three "issues" by making the output parameter optional. If not given when calling the profile option py-spy could write the output in a file called '<program_name>-<timestamp>'. (For the case of sampling an already running process we could ether use the pid instead of the program name or query the name of the executable)

Would you accept a PR for this? I would be interested in doing this as a 'first patch' to get some rust-exercise and first step into this awesome tool.

Nudin:
@benfred: *ping* – I wouldn't like to start if I don't know your interested in a PR.

benfred:
Thanks @nudin - I'm definitely interested! Sorry for the delayed response.

Can you make sure to set the extension for the filename appropriately base off the file format (like '.svg' for flamegraph, '.json' for speedscope, '.txt' for raw) ?

benfred:
Fixed by #229

PR #355, opened 2021-02-12
Make '--output' optional (#229)

Nudin:
If no filename is specified by the user it is auto generated from filename/pid, the current time and the appropriate extension.

I opened #229 a long time ago but only now found the time to implement it.


benfred:
Thanks for this!

Issue #357, opened 2021-03-11
--gil option not compatible with Python 3.8.8


Jongy:
I'm running Python 3.8.8 this way:
```
$ docker run -it --rm python:3.8.8
Python 3.8.8 (default, Feb 19 2021, 17:55:44)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>
>>>
>>>
>>> while True:
...   pass
...
```

Downloaded the latest release (0.3.4), running it:
```
$ sudo py_spy-0.3.4.data/scripts/py-spy record -r 100 -p 2603154 -o flame.svg
py-spy> Sampling process 100 times a second. Press Control-C to exit.

^C
py-spy> Stopped sampling because Control-C pressed
py-spy> Wrote flamegraph data to 'flame.svg'. Samples: 164 Errors: 0
```

it works fine.

But when passing `--gil`:
```
$ sudo py_spy-0.3.4.data/scripts/py-spy record -r 100 --gil -p 2603154 -o flame.svg
py-spy> Sampling process 100 times a second. Press Control-C to exit.

^C
py-spy> Stopped sampling because Control-C pressed
[2021-03-11T09:34:22.745908125Z ERROR inferno::flamegraph] No stack counts found
Error: Failed to write flamegraph: I/O error: No stack counts found
```

I assume something has changed in the structs of 3.8.8, because e.g 3.8.5 works fine.

I'll try to look into the differences in the structs later.

PR #358, opened 2021-03-12
Detect GIL on 3.8.8

Jongy:
I actually expected it to have some real diff :shrug:

Used `generate_bindings.py`, once with `gcc` and once with `gcc -m32` to verify `target_os="linux", target_arch="x86"`. Don't have access to other archs to test them, sadly.

Closes: #357 (tested after this branch, GIL detection works fine now)

### DIFF
--- a/src/python_bindings/mod.rs
+++ b/src/python_bindings/mod.rs
@@ -88,7 +88,7 @@ pub mod pyruntime {
                     _  => Some(788)
                 }
             },
-            Version{major: 3, minor: 8, patch: 1..=7, ..} => Some(788),
+            Version{major: 3, minor: 8, patch: 1..=8, ..} => Some(788),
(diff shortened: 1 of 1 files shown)


benfred:
Thanks! 

I really need to automate this process -

Jongy:
By the way, @benfred , I think it'd much nicer for the user experience if py-spy would exit with an error in case it doesn't know the needed offsets for GIL tracking (that is, tell the user "--gil is not support for this Python version ...").

At many points I see we have `_ => None` while others have `_ => 24`and there's also `_ => Some(1416)` for some release candidates of 3.8.0.

If I understand correctly, prior to this fix, `get_tstate_current_offset` would return `None` for 3.8.8, no? So I'd expect it to fail...

I'll look into making it exit with an error (instead of exit with 0 stacks, as it does now)

PR #361, opened 2021-03-19
Exit with an error if --gil but we failed to get necessary addrs/offsets

Jongy:
Previously we'd just emit a warning, which is less noticeable; and the result of
a --gil run where without the needed information is going to empty, anyway.

I think it's better, user-experience wise (like mentioned in https://github.com/benfred/py-spy/pull/358#issuecomment-798978091). Well at least I can vouch for it personally, I'm quite annoyed by `--gil` recordings producing empty stack files / graphs, it always takes me a short while to recall the possibility that it is simply unsupported...

About the positioning of the check - might've been nicer to place this code in `main.rs`, but it's much easier if the exit decision is made next to the failure itself. I can try moving it if you think that's better.

### DIFF
--- a/src/python_spy.rs
+++ b/src/python_spy.rs
@@ -45,6 +45,14 @@ pub struct PythonSpy {
     pub dockerized: bool
 }
 
+fn exit_if_gil(config: &Config, version: &Version, msg: &str) {
+    if config.gil_only {
+        eprintln!("Cannot detect GIL holding in version '{}' on the current platform (reason: {})", version, msg);
+        eprintln!("Please open an issue in https://github.com/benfred/py-spy with the Python version and your platform.");
+        std::process::exit(1);
+    }
+}
+
 impl PythonSpy {
     /// Constructs a new PythonSpy object.
     pub fn new(pid: Pid, config: &Config) -> Result<PythonSpy, Error> {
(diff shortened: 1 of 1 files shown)


src/python_spy.rs
```diff
@@ -45,6 +45,14 @@ pub struct PythonSpy {
     pub dockerized: bool
 }
 
+fn exit_if_gil(config: &Config, version: &Version, msg: &str) {
+    if config.gil_only {
+        eprintln!("Cannot detect GIL holding in version '{}' on the current platform (reason: {})", version, msg);
+        eprintln!("Please open an issue in https://github.com/benfred/py-spy with the Python version and your platform.");
+        std::process::exit(1);
+    }
+}
```
benfred, line 54: In the case where ```--gil``` wasn't being set, we used to produce a warning if it wasn't added. I think we should keep this warning when we're not exiting with an error:

```suggestion
    }
    warn!("Unable to detect GIL usage: {}", msg);
}
```

benfred (commented):
I think this is a good idea! thanks for adding,

src/python_spy.rs
Jongy: Yeah I thought about it, but I realized the error is just redundant because if I didn't request `--gil` I probably don't care about seeing it. But ofc it's okay to add it back :)

Issue #109, opened 2019-04-16
Cannot attach to long running process


schooft:
I get the following error
```
$ sudo py-spy --pid 9134
Error: No such file or directory (os error 2)
```

`strace` shows that this happens at 
```
open("/usr/bin/python3.5 (deleted)", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
```

The problem seems to be
```
$ sudo readlink /proc/9134/exe
/usr/bin/python3.5 (deleted)
```
`/usr/bin/python3.5` does exist but was upgraded during the runtime of the process.

benfred:
@aspin : This bug is caused by the python executable being deleted while the program is run, making it so that we can't read in information from it.  This bug will affect all versions of py-spy including 0.1.11.

Do you have a different problem? I'd be interested in hearing more about it if so

aspin:
Hmm, I'm not sure the source of the problem then. It looks something like this for me:

```
$ /usr/local/bin/py-spy -V
py-spy 0.3.0
$ /usr/local/bin/py-spy top --pid 22159
Error: No such file or directory (os error 2)
$ /usr/local/bin/py-spy -V # after installing new version
py-spy 0.1.11
$ /usr/local/bin/py-spy top --pid 22159 # works
```
```
$ sudo readlink /proc/22159/exe
/usr/local/bin/python3.7
$ /usr/local/bin/python3.7
bash: /usr/local/bin/python3.7: No such file or directory
```

For a bit of a context, the process is running inside of a Docker container, but I still seem to be able to get its PID from `ps aux`. Any helpful information I can provide?

strace output (I'm not familiar with this):
```
$ strace /usr/local/bin/py-spy top --pid 22159
execve("/usr/local/bin/py-spy", ["/usr/local/bin/py-spy", "top", "--pid", "22159"], [/* 26 vars */]) = 0
mmap(NULL, 5384, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7fab8408e000
arch_prctl(ARCH_SET_FS, 0x7fab8408f3e0) = 0
set_tid_address(0x7fab8408f418)         = 832
rt_sigaction(SIGPIPE, {SIG_IGN, [], SA_RESTORER|SA_RESTART, 0x766273}, {SIG_DFL, [], 0}, 8) = 0
[...]

Issue #360, opened 2021-03-15
Error: Failed to parse python binary; Reason: No such file or directory (os error 2)


yarikoptic:
well, my setup is quite obscure (described below) but I think it should have worked.  And if it isn't supposed to - would have been nice if error message was more informative: which binary it failed to parse and/or which file/directory is missing (I do not see some `--debug` or `--verbose` in the output of `py-spy [top] --help` which could optionally start giving me more information).

Setup: a singularity container in which I am running a python process using system wide (debian) python3.

```
(py-spy) bids@rolando:~$ export | grep SING
declare -x SINGULARITY_CONTAINER="reproin_0.9.0.simg"
declare -x SINGULARITY_NAME="reproin_0.9.0.simg"

```

 Since container environment is "fixed" and I have not pre-installed py-spy into it, I have initiated python virtualenv in it, using the same python3:
```
python3 -m virtualenv --system-site-packages --python python3 /tmp/py-spy
```

which I activated and then `pip install py-spy` into it.  Trying to run `py-spy top` though results in
```
(py-spy) bids@rolando:~$ py-spy top -p 39864
Error: Failed to parse python binary
Reason: No such file or directory (os error 2)

(py-spy) bids@rolando:~$ ls -l `which python`
lrwxrwxrwx 1 bids bids 7 Mar 15 17:36 /tmp/py-spy/bin/python -> python3
(py-spy) bids@rolando:~$ ls -l `which python3`
-rwxr-xr-x 1 bids bids 4861504 Mar 15 17:36 /tmp/py-spy/bin/python3
```

stracing it leads to

```
[pid 40252] open("/proc/39864/root/usr/local/singularity_v2.5.1/var/singularity/mnt/final/usr/bin/python3.7", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
```
so I guess it is the one it is trying to parse somehow instead of just 

```
(py-spy) bids@rolando:~$ ls -l /usr/bin/python3.7     
-rwxr-xr-x 2 root root 4861504 Jul 25  2020 /usr/bin/python3.7
```

Jongy:
My fix in #364 is relevant to 2 cases:
1. Deleted executables are accessible via `/proc/pid/exe` and not via `/proc/pid/root/...`
[...]

PR #364, opened 2021-03-20
Use /proc/pid/exe for the Python binary in Linux

Jongy:
Instead of "/proc/pid/root/$(readlink /proc/pid/exe)", we'll just use /proc/pid/exe
which provides access to the file even if it was deleted on-disk.

This is not uncommon - for example, Python may be upgraded and thus long-running
processes no longer have the correct file at the specified path.
After this change we can profile/dump those processes (only if they are Python-binary
based and not libpython-based; the latter is inaccessible if deleted... perhaps only
partially accessible via /proc/pid/map_files).

Note: this depends on https://github.com/rbspy/proc-maps/pull/8 because currently we don't extract the full filename from `/proc/pid/maps`, so the comparison in `is_python_bin` (matching `readlink("/proc/pid/exe")` to filenames in `/proc/pid/maps`) always fails.

With this PR (and the `proc-maps` fix) I managed to `py-spy dump` Python processes that were started before I upgraded Python on my local box :) (so they now run `/usr/bin/python3.8 (deleted)`)

### DIFF
--- a/src/binary_parser.rs
+++ b/src/binary_parser.rs
@@ -26,10 +26,20 @@ impl BinaryInfo {
 }
 
 /// Uses goblin to parse a binary file, returns information on symbols/bss/adjusted offset etc
-pub fn parse_binary(_pid: remoteprocess::Pid, filename: &str, addr: u64, size: u64) -> Result<BinaryInfo, Error> {
+pub fn parse_binary(_pid: remoteprocess::Pid, filename: &str, addr: u64, size: u64, _is_bin: bool) -> Result<BinaryInfo, Error> {
     // on linux the process could be running in docker, access the filename through procfs
+    // if filename is the binary executable (not libpython) - take it from /proc/pid/exe, which works
+    // across namespaces just like /proc/pid/root, and also if the file was deleted.
     #[cfg(target_os="linux")]
-    let filename = &format!("/proc/{}/root{}", _pid, filename);
+    let _tmp;
+    #[cfg(target_os="linux")]
+    let filename = if _is_bin {
+        _tmp = format!("/proc/{}/exe", _pid);
+        &_tmp
+    } else {
+        _tmp = format!("/proc/{}/root{}", _pid, filename);
+        &_tmp
+    };
 
     let offset = addr;
 
--- a/src/binary_parser.rs
+++ b/src/binary_parser.rs
@@ -32,9 +32,9 @@ pub fn parse_binary(_pid: remoteprocess::Pid, filename: &str, addr: u64, size: u
     // across namespaces just like /proc/pid/root, and also if the file was deleted.
     #[cfg(target_os="linux")]
     let filename = &if _is_bin {
-        format!("/proc/{}/exe", _pid);
+        format!("/proc/{}/exe", _pid)
     } else {
-        format!("/proc/{}/root{}", _pid, filename);
+        format!("/proc/{}/root{}", _pid, filename)
     };
 
     let offset = addr;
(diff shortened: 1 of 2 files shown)


src/binary_parser.rs
```diff
@@ -26,10 +26,20 @@ impl BinaryInfo {
...
-pub fn parse_binary(_pid: remoteprocess::Pid, filename: &str, addr: u64, size: u64) -> Result<BinaryInfo, Error> {
+pub fn parse_binary(_pid: remoteprocess::Pid, filename: &str, addr: u64, size: u64, _is_bin: bool) -> Result<BinaryInfo, Error> {
     // on linux the process could be running in docker, access the filename through procfs
+    // if filename is the binary executable (not libpython) - take it from /proc/pid/exe, which works
+    // across namespaces just like /proc/pid/root, and also if the file was deleted.
     #[cfg(target_os="linux")]
-    let filename = &format!("/proc/{}/root{}", _pid, filename);
+    let _tmp;
+    #[cfg(target_os="linux")]
+    let filename = if _is_bin {
```
Jongy, line 36: I don't like this part, too much boilerplate, but I'm not a Rust expert so I'm not sure how to fight the lifetime checker here... Would love a suggestion on how to improve this bit. (without the `_tmp`, I get the `creates a temporary which is freed while still in use` error)

benfred (approved):

benfred:
thanks for this! I think this will fix https://github.com/benfred/py-spy/issues/109 (and possibly even https://github.com/benfred/py-spy/issues/360)

src/binary_parser.rs
```diff
@@ -26,10 +26,16 @@ impl BinaryInfo {
...
+pub fn parse_binary(_pid: remoteprocess::Pid, filename: &str, addr: u64, size: u64, _is_bin: bool) -> Result<BinaryInfo, Error> {
     // on linux the process could be running in docker, access the filename through procfs
+    // if filename is the binary executable (not libpython) - take it from /proc/pid/exe, which works
+    // across namespaces just like /proc/pid/root, and also if the file was deleted.
     #[cfg(target_os="linux")]
-    let filename = &format!("/proc/{}/root{}", _pid, filename);
+    let filename = &if _is_bin {
+        format!("/proc/{}/exe", _pid);
+    } else {
+        format!("/proc/{}/root{}", _pid, filename);
```
benfred, line 37: err -sorry about the bad suggestion there.
```suggestion
        format!("/proc/{}/exe", _pid)
    } else {
        format!("/proc/{}/root{}", _pid, filename)
```

src/binary_parser.rs
```diff
@@ -26,10 +26,20 @@ impl BinaryInfo {
...
-    let filename = &format!("/proc/{}/root{}", _pid, filename);
+    let _tmp;
+    #[cfg(target_os="linux")]
+    let filename = if _is_bin {
+        _tmp = format!("/proc/{}/exe", _pid);
+        &_tmp
+    } else {
+        _tmp = format!("/proc/{}/root{}", _pid, filename);
+        &_tmp
+    };
```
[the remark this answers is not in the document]
Jongy: Wow. I sure don't understand Rust enough to have thought of this syntax.

Jongy:
> thanks for this! I think this will fix #109 (and possibly even #360)

#109 surely, yeah, that's exactly the issue.

Jongy:
@benfred note that we need to bump the version of `proc_maps` to accommodate for https://github.com/rbspy/proc-maps/pull/8

Issue #216, opened 2020-01-17
Windows: profiling inside a venv?


madig:
I'm having trouble running py-spy inside a venv on Windows 10 x64 1909, Python 3.8.1. It errors out with `Error: Failed to find python version from target process`. Admin privileges don't help. I presume the `python.exe` inside `venv\Scripts` is not what py-spy expects, so I copied over the actual EXE and DLL from the global Python38 folder. Profiling then works, but runs extremely slowly with considerable sampling lag.

Is this a known unsupported use-case?

benfred:
I've seen problems with venv on windows before https://github.com/benfred/py-spy/issues/81#issuecomment-477445626 - in that case what was happening was that python.exe in the virtualenv creates a subprocess that is the actual python process. Connecting up to the parent process fails because it doesn't contain a python interpreter.

We have an option to profile subprocesses which should support this use case. Can you try that?

madig:
```
> py-spy record -s -F -n -f speedscope --output pyspy.speedscope -- python .\bench.py
Can't get native stack traces with the ---subprocesses option on windows.

> py-spy record -s -F -f speedscope --output pyspy.speedscope -- python .\bench.py

Error: No python processes found in process 1580 or any of its subprocesses
```

Unless I'm doing it wrong?

IceTDrinker:
Can confirm I observe a similar behavior, are there plans to change the way py-spy tries to profile python in a Windows venv ?

Cheers

IceTDrinker:
Looking at how python sets things up on windows in the Python Home you have Libs/venv/scripts/nt which contains scripts that get copied during venv setup through "pyhton -m venv venv_dir"

Apparently the python.exe and pythonw.exe that are there are in fact venvlauncher.exe and venvwlauncher.exe that got renamed.

Not too sure on how to hook in venvlauncher quite yet though

Any pointers as to how one might poke in those executables to find the python interpreters ?

IceTDrinker:
The symptom I observe is that no child process is found (I printed the length of the spies hashmap with --subprocesses, len was 1) it seems children retrieval does not work or python.exe from venv (i.e. venvlauncher.exe in disguise) does not manage to launch the actual python interpreter.

Cheers

benfred:
Fixed by #370

jankap:
[...]

PR #370, opened 2021-04-07
Fix for venv in Windows 10, try several times to get the process' children

IceTDrinker:
- allows to profile venv python when using the subprocess (-s) flag

solves https://github.com/benfred/py-spy/issues/216

### DIFF
--- a/src/sampler.rs
+++ b/src/sampler.rs
@@ -88,7 +88,21 @@ impl Sampler {
 
         spies.insert(pid, PythonSpyThread::new(pid, None, &config)?);
         let process = remoteprocess::Process::new(pid)?;
-        for (childpid, parentpid) in process.child_processes()? {
+
+        // Try a few times to get the process' children to give time to the process to start (?)
+        // Allows to properly get children on Windows 10 in a venv with -s flag
+        let children = {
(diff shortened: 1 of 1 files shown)


IceTDrinker:
This is dirty, but I don't know if there is a way to wait for a Windows process to say it has children processes, improvements welcome but it does the (dirty) trick on my win 10

Cheers

IceTDrinker:
Gentle ping @benfred sorry if that's not the way to request a review, I can't seem to request them on the PR.

Cheers

benfred:
Thanks for the PR! I'm glad that you figured out what the issue is with profiling windows venvs.

I like the idea of retrying if we can't find any python processes - but my one issue with this PR is that it will introduce a 1 second delay in starting sampling for the case where the main process is a python process and there aren't any children yet. 
I'll fix this up in a future PR though =)

benfred:
Made that tweak here https://github.com/benfred/py-spy/pull/377

IceTDrinker:
Thanks for the merge and the tweak, I was thinking the delay might be a problem but did not think about passing the -s flag on a program with an already running python interpreter !

Cheers

PR #393, opened 2021-05-13
Document Python 3.9 support

adamchainz:
CI is testing 3.9.x, and releases since 0.3.4 have claimed some level of support, so it looks like Python 3.9 is supported.


benfred:
Thanks!  Python 3.9 is indeed fully supported =)

Issue #294, opened 2020-09-25
Specify timeunit in speedscope profiles


benfred:
Specifying the timeunit will make speedscope profiles much more readable. See https://github.com/jlfwong/speedscope/issues/311

PR #401, opened 2021-06-07
Specify speedscope units

mindriot101:
We use the sampling rate from the command line arguments to output the correct time units, so that speedscope profiles will show correct time units.

Note: we cannot use static compile time generics to provide the `Write` trait usage, because `Recorder` is used as a trait object, and as such cannot include generic parameters.

See #294


benfred (approved):
looks great! thanks for doing this,

PR #420, opened 2021-07-14
Build wheels with maturin

### LINKED ISSUE #419: No Apple Silicon wheel available
It would be ideal if wheels could be provided for machines running apple silicon processors.  I'm able to build with cargo, but projects like python `ray` provide py-spy as a dependency, which requires them to be pip-installable on the local architecture.

messense:
Closes #419 , this adds macOS universal2 wheel and source distribution, ~~though it looses Python 2 support, it's easy to add support for it in maturin but [the project leader doesn't want to support Python 2 anymore since it's EOL](https://github.com/PyO3/maturin/pull/404#issuecomment-777022619)~~, added back Python 2 support by renaming wheels.

https://github.com/PyO3/maturin


benfred (commented):
Thanks for this! This looks super interesting - and having apple silicon wheels will be useful.

I do have some questions here:

1) For python2 support, it seems like all thats necessary here is renaming the wheel files (like replacing ```py3``` with ```py2.py3```) to make this work. Even without maturin supporting python2 - it seems like we could hack in python2 support with running something like 

```
  - name: RenameWheels
      run: |
        python -c "import shutil; import glob; wheels = glob.glob('dist/*.whl'); [shutil.move(wheel, wheel.replace('py3', 'py2.py3')) for wheel in wheels if 'py2' not in wheel]"
``` 
  before the upload step to get python2 support back? I don't really want to drop python2 just yet, even though its well out of date by now. (note I haven't tested this but think conceptually something like this should work out ok).

2) The generated wheels look like they are using [pep600](https://www.python.org/dev/peps/pep-0600/) manylinux tags - do you think this will cause any problems? Like the ARM wheels are claiming we need glibc 2.17 even though they are compiled with musl and don't link glibc at all 

3) Does maturin pick up and use the version info from Cargo.toml? I just want to verify that we don't need to set in pyproject.toml too. 

4) Having apple silicon wheels is great! I'm wondering if you know of any way to get these tested during CI =) (I'm thinking we might have to wait until github actions supports this, but thought I'd ask anyways).

messense:
> 1. For python2 support, it seems like all thats necessary here is renaming the wheel files (like replacing `py3` with `py2.py3`) to make this work. Even without maturin supporting python2 - it seems like we could hack in python2 support with running something like
> 
> ```
>   - name: RenameWheels
>       run: |
>         python -c "import shutil; import glob; wheels = glob.glob('dist/*.whl'); [shutil.move(wheel, wheel.replace('py3', 'py2.py3')) for wheel in wheels if 'py2' not in wheel]"
> ```

I'll give this a try.

> 1. The generated wheels look like they are using [pep600](https://www.python.org/dev/peps/pep-0600/) manylinux tags - do you think this will cause any problems? Like the ARM wheels are claiming we need glibc 2.17 even though they are compiled with musl and don't link glibc at all

manylinux_2_17 is just an alias of manylinux2014 with no difference

1. https://github.com/pypa/auditwheel/blob/b0762a910ef159242d47f601f4605053f23b8e92/auditwheel/policy/manylinux-policy.json#L61-L62
2. https://github.com/pypa/auditwheel/pull/288


> 2. Does maturin pick up and use the version info from Cargo.toml? I just want to verify that we don't need to set in pyproject.toml too.

Yes, maturin merges metadata from `Cargo.toml` and `pyproject.toml`.

> 3. Having apple silicon wheels is great! I'm wondering if you know of any way to get these tested during CI =) (I'm thinking we might have to wait until github actions supports this, but thought I'd ask anyways).

The only way to test the arm64 part of it on CI now is using a self hosted runner.

messense:
I've verified it works on arm64 macOS.

```bash
❯ ls
py_spy-0.3.7-py2.py3-none-macosx_10_7_x86_64.whl                                          py_spy-0.3.7-py2.py3-none-manylinux_2_17_armv7l.manylinux2014_armv7l.whl                  py_spy-0.3.7-py2.py3-none-win_amd64.whl
py_spy-0.3.7-py2.py3-none-macosx_10_9_x86_64.macosx_11_0_arm64.macosx_10_9_universal2.whl py_spy-0.3.7-py2.py3-none-manylinux_2_5_i686.manylinux1_i686.whl                          py_spy-0.3.7.tar.gz
py_spy-0.3.7-py2.py3-none-manylinux_2_17_aarch64.manylinux2014_aarch64.whl                py_spy-0.3.7-py2.py3-none-manylinux_2_5_x86_64.manylinux1_x86_64.whl

messense in messenses-MBP in ~/Downloads/py-spy via 🐍 v3.9.2 (test)
❯ uname -a
Darwin messenses-MBP.lan 20.5.0 Darwin Kernel Version 20.5.0: Sat May  8 05:10:31 PDT 2021; root:xnu-7195.121.3~9/RELEASE_ARM64_T8101 arm64

messense in messenses-MBP in ~/Downloads/py-spy via 🐍 v3.9.2 (test)
❯ pip install py_spy-0.3.7-py2.py3-none-macosx_10_9_x86_64.macosx_11_0_arm64.macosx_10_9_universal2.whl
Processing ./py_spy-0.3.7-py2.py3-none-macosx_10_9_x86_64.macosx_11_0_arm64.macosx_10_9_universal2.whl
Installing collected packages: py-spy
Successfully installed py-spy-0.3.7

messense in messenses-MBP in ~/Downloads/py-spy via 🐍 v3.9.2 (test)
❯ py-spy --help
py-spy 0.3.7
Sampling profiler for Python programs

USAGE:
    py-spy <SUBCOMMAND>

OPTIONS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    record    Records stack trace information to a flamegraph, speedscope or raw file
    top       Displays a top like view of functions consuming CPU
    dump      Dumps stack traces for a target program to stdout
    help      Prints this message or the help of the given subcommand(s)

❯ file ~/.pyenv/versions/test/bin/py-spy
/Users/messense/.pyenv/versions/test/bin/py-spy: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64]
/Users/messense/.pyenv/versions/test/bin/py-spy (for architecture x86_64):	Mach-O 64-bit executable x86_64
/Users/messense/.pyenv/versions/test/bin/py-spy (for architecture arm64):	Mach-O 64-bit executable arm64
```

benfred (approved):
This is great! 

Thanks for making this work with python2 and verifying that this works w/ arm64 osx,

PR #421, opened 2021-07-14
Add --syscalls option (linux only)

manuels:
This commit adds a new option (--syscalls/-S) to trace what time your python programm spends in kernel space.
It is only supported on linux where current syscall can be determined using /proc/{pid}/task/{tid}/syscall.


manuels:
Example output of top:
```
  %Own   %Total  OwnTime  TotalTime  Function (filename:line)                                   
 99.00%  99.00%   67.60s    67.60s   select syscall (<kernel>)
  1.00%   1.00%   0.200s    0.200s   write syscall (<kernel>)
  0.00%   1.00%   0.130s    0.390s   x (x.py:5)
  0.00%   0.00%   0.060s    0.060s   brk syscall (<kernel>)
  0.00%   0.00%   0.010s    0.010s   x (x.py:3)
  0.00%  99.00%   0.000s    67.60s   x (x.py:4)
  0.00% 100.00%   0.000s    68.00s   <module> (x.py:7)
```

Jongy:
Nice!

I had something similar in mind a while back - to add `/proc/pid/stack` entries as traces (so you have kernel stacks in addition to the Python stacks). It's simpler (because the kernel does symbolication for you) but requires root IIRC (unlike `/proc/pid/syscall`). 

Also, I'm not sure it's useful unless you are in `--nonblocking` mode - otherwise, the kernel stack would be "stopped in ptrace".

manuels:
Yes, my first idea was to add `/proc/pid/stack`, too, but then I thought that a programmer that only sees the python stacktrace (and not the c stacktrace of cpython) is probably barely interested in the internals of the kernel ;)

src/config.rs
```diff
@@ -282,6 +297,13 @@ impl Config {
         config.include_idle = matches.occurrences_of("idle") > 0;
         config.gil_only = matches.occurrences_of("gil") > 0;
         config.include_thread_ids = matches.occurrences_of("threads") > 0;
+        #[cfg(all(target_os = "linux", feature = "trace_syscalls"))]
+        {
+            config.trace_syscalls = matches.occurrences_of("trace_syscalls") > 0;
+            if config.trace_syscalls {
+                config.include_idle = true;
+            }
```
benfred, line 305: Why do we need to include idle frames when we're including syscalls?

src/python_spy.rs
```diff
@@ -17,6 +17,400 @@ use remoteprocess::{Process, ProcessMemory, Pid, Tid};
 use proc_maps::{get_process_maps, MapRange};
 
 
+#[cfg(all(target_os = "linux", feature = "trace_syscalls"))]
+lazy_static! {
+    static ref SYSCALLS: HashMap<u32, &'static str> = {
+        let mut m = HashMap::new();
```
benfred, line 23: How was the list of syscalls generated? What did you use as a datasource here?

benfred (commented):
Thanks for the PR!

src/python_spy.rs
manuels: I extracted it from https://github.com/torvalds/linux/blob/9b76d71fa8be8c52dbc855ab516754f0c93e2980/arch/x86/entry/syscalls/syscall_64.tbl, but maybe it makes sense to support multiple architectures here or fail if the architecture is not x86_64

src/config.rs
manuels: That's a good question: if you call `time.sleep(1)` python will effectively run the `select()` syscall. What should happen? Should py-spy show the select syscall or should it ignore the thread because effectively it's idle?
Maybe you are right that this decision should be taken care of by the user, so --syscalls should not include --idle (only if the user specifies this argument).
I just wanted to bring the problem with `time.sleep()` to your attention, but probably --syscalls should not necessarily inlcude --idle

manuels: I changed this, --syscalls does not imply --idle

src/python_spy.rs
manuels: I changed this to use the `syscalls` crate to determine the name of the syscalls

manuels:
Hey, @benfred, let me know if there is anything in this pull request that I should change if you are not happy with the current state. :D

manuels:
Hey, @benfred is there anything else I can do to have this merged?

src/config.rs
```diff
@@ -249,6 +255,14 @@ impl Config {
         #[cfg(unwind)]
         let dump = dump.arg(native.clone());
 
+        // add syscall traces if appropiate
```
kapilt, line 258: typo on appropriate

manuels: Fixed

manuels:
Hey, @benfred, maybe you missed my last message. Is there anything else I can do to have this merged?

PR #427, opened 2021-07-31
Add a subcommand to generate shell completions

messense:
This can be used by Homebrew formula to automatically generate shell completions. 

```bash
$ py-spy completions --help
py-spy-completions
Generate shell completions

USAGE:
    py-spy completions [shell]

OPTIONS:
    -h, --help       Prints help information
    -V, --version    Prints version information

ARGS:
    <shell>    Shell type [possible values: zsh, bash, fish, powershell, elvish]
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,4 +1,4 @@
-use clap::{App, Arg, crate_description, crate_name, crate_version, arg_enum, value_t};
+use clap::{App, AppSettings, Arg, crate_description, crate_name, crate_version, arg_enum, value_t};
 use remoteprocess::Pid;
 
 /// Options on how to collect samples from a python process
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -269,6 +277,11 @@ impl Config {
             "top" => {
                 config.sampling_rate = value_t!(matches, "rate", u64)?;
             }
+            "completions" => {
+                let shell = value_t!(matches.value_of("shell"), clap::Shell).unwrap_or_else(|e| e.exit());
+                app.gen_completions_to(crate_name!(), shell, &mut std::io::stdout());
+                std::process::exit(0);
```
messense, line 283: I don't like this, but IMO `Config` is currently too tied with `clap::App`.

messense: In case you don't know, I'd recommend using [structopt](https://github.com/TeXitoi/structopt) instead of using `clap` directly.

benfred:
thanks!

PR #434, opened 2021-08-10
Fix speedscope time interval

### LINKED ISSUE #430: Time scale off when exporting to speedscope
Example program:
```
import time

def do_something():
    time.sleep(0.1)
    print('wake')

def main():
    while True:
        time.sleep(0.9)
        do_something()

if __name__ == '__main__':
    main()
```
Run py-spy with: 
`py-spy record -f speedscope -i -d 5 -p 2311089`

When the output is imported into speedscope, it looks like the absolute times are off by 100x.

gjoseph92:
This fixes two things:
1. With multiple threads/processes, each may be a different length, so the `end_value` needs to be set independently. (Also before, the `end_value` was mistakenly the number of threads, not the number of samples in a thread.)
2. `weights` are (apparently) not unitless, and need to be given in seconds.

Closes #430

Didn't have time to add tests, but I could if you'd like.

FWIW I was manually testing with:
```python
import threading
import time

threads = [threading.Thread(target=time.sleep, args=(i / 2,)) for i in range(4)]
for t in threads:
    t.start()

for t in threads:
    t.join()
```

### DIFF
--- a/src/speedscope.rs
+++ b/src/speedscope.rs
@@ -127,10 +127,6 @@ enum ValueUnit {
 
 impl SpeedscopeFile {
   pub fn new(samples: &HashMap<Tid, Vec<Vec<usize>>>, frames: &Vec<Frame>, thread_name_map: &HashMap<Tid, String>, sample_rate: u64) -> SpeedscopeFile {
-    let end_value = samples.len();
-
-    // we sample at 100 Hz, so scale the end value to match
-    let scaled_end_value = end_value as f64 / sample_rate as f64;
     SpeedscopeFile {
       // This is always the same
       schema: "https://www.speedscope.app/file-format-schema.json".to_string(),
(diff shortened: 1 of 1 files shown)


benfred:
looks good - thanks! I tested myself before/after and the output makes a bunch more sense now 

(btw - keep up the good work on dask! I use dask for work and love it =).

gjoseph92:
Keep up the good work on py-spy! It's been invaluable lately. We know you're busy but we'd love to nerd-snipe you on https://github.com/dask/distributed/discussions/4825 and https://github.com/dask/distributed/issues/4987 someday :)

gjoseph92:
Hey @benfred whenever you have the time, it would help us a ton if you could cut a new py-spy release with this change. I'm profiling multithreaded things for dask, and having the nonsensical times is a little annoying.

benfred:
@gjoseph92 I've tagged a new release - which will be pushed out to pypi/cargo etc with by this CI run https://github.com/benfred/py-spy/actions/runs/1216029122 when it finishes

gjoseph92:
Thank you!!

PR #453, opened 2021-09-28
Fix typos discovered by codespell

cclauss:
[`codespell --ignore-words-list=als,crate,nd`](https://pypi.org/project/codespell/)

### DIFF
--- a/ci/testdata/cython_test.c
+++ b/ci/testdata/cython_test.c
@@ -5114,7 +5114,7 @@ static CYTHON_INLINE int __Pyx_PyErr_GivenExceptionMatches2(PyObject *err, PyObj
     if (ctversion[0] != rtversion[0] || ctversion[2] != rtversion[2]) {
         char message[200];
         PyOS_snprintf(message, sizeof(message),
-                      "compiletime version %s of module '%.100s' "
+                      "compile time version %s of module '%.100s' "
                       "does not match runtime version %s",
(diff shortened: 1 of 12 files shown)


ci/testdata/cython_test.c
```diff
@@ -5114,7 +5114,7 @@ static CYTHON_INLINE int __Pyx_PyErr_GivenExceptionMatches2(PyObject *err, PyObj
     if (ctversion[0] != rtversion[0] || ctversion[2] != rtversion[2]) {
         char message[200];
         PyOS_snprintf(message, sizeof(message),
-                      "compiletime version %s of module '%.100s' "
+                      "compile time version %s of module '%.100s' "
```
cclauss, line 5117: Or ___compile-time___ according to https://en.wikipedia.org/wiki/Compile_time

cclauss, line 5117:
```suggestion
                      "compile-time version %s of module '%.100s' "
```
would also be acceptable according to https://en.wikipedia.org/wiki/Compile_time

benfred: this is generated code - we can fix here, but this really originates upstream in cython:

https://github.com/cython/cython/blob/e2a23fe11063ee13c886e89007207a07a397e016/Cython/Utility/ModuleSetupCode.c#L1397-L1400

benfred (approved):

PR #477, opened 2022-01-27
Update longsleep.py

yash1120:
remove blank lines


benfred:
the blank lines are from formatting with black -

PR #528, opened 2022-10-07
add callgrind output format

### LINKED ISSUE #3: callgrind output?
Please consider providing profiling output in the callgrind format, as there already exists excellent visualizers (e.g., kcachegrind) for it.  (I'm writing this same issue to each and every python profiler project that I run into and that appears interesting :-))

EpigeneMax:
Closes #3 

Following flamegraph and speedscope formats, callgrind formatting code is adapted from rbspy.
Uses Frame instead of rbspy's StackFrame.

### DIFF
--- a/src/callgrind.rs
+++ b/src/callgrind.rs
@@ -1,0 +1,473 @@
+// This code is adapted from rbspy:
+// https://github.com/rbspy/rbspy/tree/master/src/ui/callgrind.rs
+// licensed under the MIT License:
+/*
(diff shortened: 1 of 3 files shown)


rmcavoy:
Any plans to update this merge?

EpigeneMax:
I can work out the conflicts with `master`, but I think a review would be in order before merging.
@benfred would you have some bandwidth to take a look?

PR #543, opened 2022-12-11
Test Windows Wheels and Fix symbol address resolution on Windows

benfred:
This adds CI for testing out windows wheels against various versions of python.

This also applies a fix to use RVA for symbol address resolution on windows (as identified by @akhramov in #672)


benfred:
the remaining windows failures should be fixed w/ https://github.com/benfred/py-spy/pull/718

PR #554, opened 2023-02-21
Fix codestyle inconsistencies (cargo fmt)

danog:
Also added a CI step that runs `cargo fmt --check`.


benfred:
Thanks for the PR!  There are a lot of changes to review here though - and while I doubt that you've changed anything beyond just running `cargo fmt` on the code,  it's easier for me to just re-apply cargo fmt myself (which I've done as part of #590).

PR #555, opened 2023-02-21
Implement Pyroscope reporting

danog:
Based on #554, implements [Pyroscope](https://pyroscope.io) reporting directly inside py-spy.


danog:
Ping? :)

benfred (commented):
This might be more of a questions for @petethepig - but I'm wondering where this PR fits in with the pyroscope roadmap.

I saw that pyroscope was recently acquired by Grafana recently (congratulations on this btw!),  and I'm wondering what the plans are moving forward for using py-spy with pyroscope. There used to be a pyroscope project https://github.com/pyroscope-io/py-spy for integration - but I see that project is now archived.

PR #576, opened 2023-05-18
Fix freebsd builds

madkinsz:
Fixes builds for freebsd

- Update freebsd build runner to `macos-latest`
- Use a newer version of llvm (previous version is not found)

See failure at https://github.com/benfred/py-spy/actions/runs/4984539891/jobs/8946184710
Will require https://github.com/benfred/py-spy/pull/574
View successful build at https://github.com/madkinsz/py-spy/pull/7


benfred (approved):
thanks for this!

PR #578, opened 2023-05-20
Upload a sdist package to pypi

benfred:


madkinsz:
The builds are fixed in https://github.com/benfred/py-spy/pull/574

PR #626, opened 2023-10-26
Page-align offset before calculating virtual addresses for ELF

andrewjcg:
This mimics what `ld.so` does by aligning the virtual addresses to the current page size, and fixes:
- Existing issues with DSOs produced by LLD which, unlike gold or BFD, doesn't align p_vaddr to p_align (the previous overflow prevention wasn't sufficient here).
- Systems w/ configurable page sizes, where just using p_align isn't enough.


src/binary_parser.rs
```diff
@@ -150,9 +150,11 @@ pub fn parse_binary(filename: &Path, addr: u64, size: u64) -> Result<BinaryInfo,
...
                 })?;
 
-            // p_vaddr may be larger than the map address in case when the header has an offset and
-            // the map address is relatively small. In this case we can default to 0.
-            let offset = offset.saturating_sub(program_header.p_vaddr);
+            // Align the virtual address offset, then subtract it from the offset
+            // to get real offset for symbol addresses in the file.
+            let aligned_vaddr =
+                program_header.p_vaddr - (program_header.p_vaddr % page_size::get() as u64);
+            let offset = offset - aligned_vaddr;
```
benfred, line 157:
```suggestion
            let offset = offset.saturating_sub(aligned_vaddr);
```

PR #628, opened 2023-10-31
Handle trailing build metadata in Python version string

andrewjcg:
Don't fail Python versions when the version string contains trailing build metadata after the `+` (as per: https://semver.org/#spec-item-10).


benfred (commented):
Can you add a test case that exercises this functionality? Something like https://github.com/benfred/py-spy/blob/75adc45e6323e654ec21780bec60b182e2980b7b/src/version.rs#L132-L143

andrewjcg:
Added test and also updated regex to match full build metadata format supported by https://semver.org/#spec-item-10.

PR #630, opened 2023-11-07
Speed up --gil by skipping work for non-GIL threads

wmanley:
When the user has specified --gil we needn't collect stack traces for the non-GIL threads.  This is a significant speedup on heavily threaded Python code.

My server has 271 threads (most of which are idle).

With the previous code (sampling at 10Hz) I get:

    $ ./py-spy record -o /var/opt/stbt/profile.svg --pid $(pgrep -f http_service.service) -r 10 -d 10 --gil
    py-spy> Sampling process 10 times a second for 10 seconds. Press Control-C to exit.

    py-spy> 1.29s behind in sampling, results may be inaccurate. Try reducing the sampling rate
    py-spy> 1.36s behind in sampling, results may be inaccurate. Try reducing the sampling rate
    py-spy> 1.32s behind in sampling, results may be inaccurate. Try reducing the sampling rate
    py-spy> 1.22s behind in sampling, results may be inaccurate. Try reducing the sampling rate
    py-spy> 1.00s behind in sampling, results may be inaccurate. Try reducing the sampling rate
    py-spy> Wrote flamegraph data to '/var/opt/stbt/profile.svg'. Samples: 3 Errors: 51

With this commit I can sample at 80Hz, and the error rate is way lower:

    $ ./py-spy2 record -o /var/opt/stbt/profile.svg --pid $(pgrep -f http_service.service) -r 80 -d 10 --gil
    py-spy> Sampling process 80 times a second for 10 seconds. Press Control-C to exit.

    py-spy> 1.00s behind in sampling, results may be inaccurate. Try reducing the sampling rate
    py-spy> Wrote flamegraph data to '/var/opt/stbt/profile.svg'. Samples: 43 Errors: 0

I suspect the difference in error rate is due to the error handling in `_get_stack_traces` - an error getting information about any thread means that all the traces are thrown away.

### DIFF
--- a/src/python_spy.rs
+++ b/src/python_spy.rs
@@ -189,9 +189,14 @@ impl PythonSpy {
     fn _get_stack_traces<I: InterpreterState>(&mut self) -> Result<Vec<StackTrace>, Error> {
         // Query the OS to get if each thread in the process is running or not
         let mut thread_activity = HashMap::new();
-        for thread in self.process.threads()?.iter() {
-            let threadid: Tid = thread.id()?;
-            thread_activity.insert(threadid, thread.active()?);
+        if self.config.gil_only {
+            // Don't need to collect thread activity if we're only getting the
+            // GIL thread: If we're holding the GIL we're by definition active.
+        } else {
+            for thread in self.process.threads()?.iter() {
+                let threadid: Tid = thread.id()?;
+                thread_activity.insert(threadid, thread.active()?);
+            }
         }
 
         // Lock the process if appropriate. Note we have to lock AFTER getting the thread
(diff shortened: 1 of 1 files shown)


src/python_spy.rs
```diff
@@ -128,56 +128,56 @@ impl PythonSpy {
     }
 
     /// Gets a StackTrace for each thread in the current process
-    pub fn get_stack_traces(&mut self) -> Result<Vec<StackTrace>, Error> {
+    pub fn get_stack_traces(&mut self, gil_only: bool) -> Result<Vec<StackTrace>, Error> {
```
benfred, line 131: Rather than pass the `gil_only` flag as a parameter here - can we do the same as other config options and just read from the `self.config.gil_only` member ? 


```suggestion
    pub fn get_stack_traces(&mut self) -> Result<Vec<StackTrace>, Error> {
```

benfred (changes requested):
thanks for the PR!

src/python_spy.rs
wmanley: Done. It's a much smaller diff now :)

benfred (approved):
thanks!

PR #631, opened 2023-11-09
Support facebook's cinder.3.8 Python fork

andrewjcg:
Facebook's cinder.3.8 Python fork uses a different offset for the current tstate.  Use the buildmetadata to detect and special case this (https://github.com/facebookincubator/cinder/blob/cinder/3.8/Include/patchlevel.h#L26).


benfred (approved):
lgtm

Issue #617, opened 2023-10-01
Failing to get stacktrace for native on Linux


fzakaria:
Hi!

I am trying to use py-spy on my project [sqlelf](https://github.com/fzakaria/sqlelf).
It's failing to generate native stack traces through with the `--native` flag turned on.

I enabled logging and see the following (repeating):
```
[2023-10-01T20:07:50.036807699Z DEBUG remoteprocess::linux] attached to thread 86516
[2023-10-01T20:07:50.037880260Z DEBUG remoteprocess::linux] detached from thread 86516
[2023-10-01T20:07:50.037933250Z WARN  py_spy] Failed to get stack trace from 86516: Failed to merge native and python frames (Have 3 native and 8 python)
[2023-10-01T20:07:50.038235706Z DEBUG remoteprocess::linux] attached to thread 86516
[2023-10-01T20:07:50.039267936Z DEBUG remoteprocess::linux] detached from thread 86516
[2023-10-01T20:07:50.039317932Z WARN  py_spy] Failed to get stack trace from 86516: Failed to merge native and python frames (Have 4 native and 9 python)
[2023-10-01T20:07:50.040563003Z DEBUG remoteprocess::linux] attached to thread 86516
[2023-10-01T20:07:50.041708051Z DEBUG remoteprocess::linux] detached from thread 86516
[2023-10-01T20:07:50.041758075Z WARN  py_spy] Failed to get stack trace from 86516: Failed to merge native and python frames (Have 3 native and 8 python)
```

What can I do to debug this?

```
❯ py-spy --version
py-spy 0.3.14

❯ python --version
Python 3.11.5

❯ uname -r
6.4.4-3rodete1-amd64
```

Issue #558, opened 2023-03-16
python311 do not support --native 


1995chen:
py-spy record -o grpc.svg --pid 267 --native --subprocesses --idle --gil -- --cap-add SYS_PTRACE
Error: No python processes found in process 267 or any of its subprocesses
(py_311) root@delta-Precision-5820-Tower:/workspace/project# py-spy record -o grpc.svg --pid 267 --subprocesses --idle --gil -- --cap-add SYS_PTRACE
py-spy> Sampling process 100 times a second. Press Control-C to exit.

DHolmanCoding:
Yeah, full 3.11 would be amazing!

MorningLightMountain713:
3.11 support +1

PR #635, opened 2023-11-16
Fix Python 3.11 native sampling

### LINKED ISSUE #634: Python 3.11 --native profiling fails because of bpo-45256
[bpo-45256](https://github.com/python/cpython/issues/89419) removed the 1-to-1 mapping between the CPython stack frames (`_PyEval_EvalFrame`) and Python stack frames.

This breaks the `--native` extension for Python 3.11. In particular, even the simplest call stacks can't be merged anymore:

```
import time
import os

def wait(sleep_time):
    time.sleep(sleep_time)

print("PID", os.getpid())
wait(30)
```

```
Process 624584: python wait.py
Python v3.11.5 (...)

Thread 624584 (idle)
    wait (wait.py:5)
    <module> (wait.py:8)

> py-spy dump --pid 624584 -n
Process 624584: python wait.py
Python v3.11.5 (...)

Error: Failed to merge native and python frames (Have 1 native and 2 python)
```

In previous Python versions (until 3.10.x), each line in the python stack trace corresponded to one `_PyEval_EvalFrame` (and `_PyEval_EvalFrameDefault`) frame in CPython, but this is not the case anymore. For a program like the one above, there is only one `_PyEval_EvalFrame`.

Unfortunately, it doesn't seem straightforward to fix this. In the simplest cases (like here), where there's only 1 native frame, we could probably just insert the full python stack trace at that location. However, in more complicated situations (e.g. when calling a native extension, which then executed python code), this wouldn't suffice.

The challenge is to identify which groups of python frames belong to which `_PyEval_EvalFrame`. I've digged a bit into this, but couldn't come up with a good way so far.

Note that this issue has been raised previously, e.g. in https://github.com/benfred/py-spy/issues/617, https://github.com/benfred/py-spy/issues/608, https://github.com/benfred/py-spy/issues/558. I'm opening this issue to consolidate these with some relevant background.

I'm happy to contribute, but my knowledge of CPython internals is limited, so a few pointers would be appreciated.

krfricke:
In Python 3.11, merging native frames works a bit differently. We have to detect if a frame is an "entry frame". Once we hit a native frame that corresponds to python, we need to keep merging python frames until we hit the entry frame.

pystack does it very similarly: https://sourcegraph.com/github.com/bloomberg/pystack/-/blob/src/pystack/_pystack/pyframe.cpp?L73-104

Confirmed this works for me in a "sandwiched" environment, where Python calls Rust, which calls Python again, as well as the simple example from #634.

Closes #634, #617, presumably also #558

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -284,6 +284,7 @@ fn record_samples(pid: remoteprocess::Pid, config: &Config) -> Result<(), Error>
                     short_filename: None,
                     line: 0,
                     locals: None,
+                    is_entry: true,
                 });
             }
 
--- a/src/native_stack_trace.rs
+++ b/src/native_stack_trace.rs
@@ -93,8 +93,18 @@ impl NativeStack {
                         // if we have a corresponding python frame for the evalframe
                         // merge it into the stack. (if we're out of bounds a later
                         // check will pick up - and report overall totals mismatch)
-                        if python_frame_index < frames.len() {
+
+                        // Merge all python frames until we hit one with `is_entry`.
+                        let mut is_entry = false;
+                        while !is_entry && python_frame_index < frames.len() {
+                            is_entry = frames[python_frame_index].is_entry;
                             merged.push(frames[python_frame_index].clone());
+
+                            // We always increase `python_frame_index` outside the loop,
+                            // so inside we should only increase when we don't continue.
+                            if !is_entry {
+                                python_frame_index += 1;
+                            }
                         }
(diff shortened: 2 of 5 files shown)


benfred (approved):
thanks for diving into this and coming up with the fix - nice work!

The change looks good to me

I've created a follow up issue here to track adding tests for the native sampling to CI https://github.com/benfred/py-spy/issues/640 -

PR #636, opened 2023-11-16
Re-fetch thread ID in native profiling for Python 3.11

krfricke:
Python 3.11 exposes the native thread ID in the thread state.

However, it seems this can be out of sync/stale when a process forked in a native extension. This lead to errors such as:

```
Process 53014: ...
Python v3.11.5 (/root/.pyenv/versions/3.11.5/bin/python3.11)

Error: UNW_EBADREG: bad register number
Reason: UNW_EBADREG: bad register number
```

Upon investigation, the native thread ID (which in this case is just the PID) was still pointing to the parent process PID.

The easiest fix here is to just use the existing logic in py-spy to retrieve the thread ID from the OS. This leads to the desired result:

```
Process 53014: ...
Python v3.11.5 (...)

Thread 53014 (idle): "MainThread"
...
```

I'm not sure if this is a bug on the Python side - I can see why they wouldn't update/poll a new thread ID after a fork in a native extension - afaik there is no way for Python to tell it's been forked, and it's likely set on init. I haven't checked the CPython source for this though. I might investigate further, but since this fix resolves my problem, won't dive too deep into it.

### DIFF
--- a/src/python_spy.rs
+++ b/src/python_spy.rs
@@ -249,8 +249,11 @@ impl PythonSpy {
 
             // python 3.11+ has the native thread id directly on the PyThreadState object,
             // for older versions of python, try using OS specific code to get the native
-            // thread id (doesn't work on freebsd, or on arm/i686 processors on linux)
-            if trace.os_thread_id.is_none() {
+            // thread id (doesn't work on freebsd, or on arm/i686 processors on linux).
+            // On native profiling of forked processes where the fork happens in native code,
+            // Python 3.11+'s native thread ID is not always updated. In that case, also re-fetch
+            // the native thread ID from the OS.
+            if trace.os_thread_id.is_none() || self.config.native {
                 let mut os_thread_id = self._get_os_thread_id(python_thread_id, &interp)?;
 
                 // linux can see issues where pthread_ids get recycled for new OS threads,


benfred:
It seems like you're hitting a bug in cpython https://github.com/python/cpython/issues/100649 - does the fix included there resolve this issue for you?

I'd rather not go back to the previous methods of detecting OS thread ids - being able to directly get the OS threadid from the python interpreter is definitely the way to go here (the previous method doesn't work on ARM for instance).

krfricke:
The fix is included in Python 3.11.2 (https://docs.python.org/release/3.11.2/whatsnew/changelog.html) but I still run into it in 3.11.5 (as per this issue).

I can try to come up with an easily reproducible example.

Is there a way to force the python interpreter to refresh its thread ID? I could potentially do that after a fork in native code.

benfred:
Are you calling `PyOS_AfterFork_Child`  in your c-extension? I don't think the fix will get included without that. https://docs.python.org/3/c-api/init.html#cautions-about-fork

Also according to https://docs.python.org/3/library/os.html#os.register_at_fork :
> Note that fork() calls made by third-party C code may not call those functions, unless it explicitly calls [PyOS_BeforeFork()](https://docs.python.org/3/c-api/sys.html#c.PyOS_BeforeFork), [PyOS_AfterFork_Parent()](https://docs.python.org/3/c-api/sys.html#c.PyOS_AfterFork_Parent) and [PyOS_AfterFork_Child()](https://docs.python.org/3/c-api/sys.html#c.PyOS_AfterFork_Child).

krfricke:
Apologies for the delay, last week was a bit busy.

It looks like calling `PyOS_AfterFork_Child` solves this problem - thanks for the help, let's discard this PR then.

PR #637, opened 2023-11-16
Add `--native-all` flag that also print the stack of non-python threads

krfricke:
Similar to pystack's `--native-all`, this PR adds a flag with the same name to also print threads that may have been spawned from native extensions (but don't have an associated python interpreter).

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -173,7 +173,7 @@ impl Config {
         #[cfg(unwind)]
             let native_all = Arg::new("native-all")
             .short('N')
-            .long("native=all")
+            .long("native-all")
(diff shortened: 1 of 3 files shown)


krfricke:
@benfred I'm just going through some of my open PRs, is there any interest in merging this?

d4l3k:
This would be handy -- running into an issue where we have mixed threads from Python and rust and `--native` isn't working

krfricke:
@d4l3k the problem you're describing is likely resolved by https://github.com/benfred/py-spy/pull/751

pdpino:
Are there any plans to merge this in the near future? Just wondering because this would be super useful!

PR #648, opened 2024-01-04
Add support for UCS-2 strings 

nariman:
Trying to fix an issue, when py-spy would not parse a stack trace containing functions with cyrillic names.

https://github.com/benfred/py-spy/blob/4a11c44849aa2146634ceef8ee8eb43bc1e969ad/src/python_data_access.rs#L26-L31

Seems like UCS-2 strings are in use by _some_ (maybe all?) pre-built Python binaries (tested on `mcr.microsoft.com/devcontainers/rust:1-1-bullseye` Dev Container, our company's CentOS environment, as well as my local environment with Python 3.7 and 3.11 versions installed using pyenv).

If there's a function with a UCS-2 encoded name on the first recorded stack trace, py-spy just fails.

<details>
<summary>Example</summary>

```py
import time

def кириллица(seconds):
    time.sleep(seconds)

if __name__ == "__main__":
    кириллица(10)
```

Outputs (`profile.svg` will not be created in this case):

```sh
$ py-spy record -o profile.svg -- python3 tests/scripts/cyrillic.py 
Error: Failed to find a python interpreter in the .data section
```
</details>

If py-spy successfully started the recorder, we'll get a flamegraph built without stack traces with UCS-2 strings, so final data/graph will look strange or even misleading.

<details>
<summary>Example</summary>

```py
import time

def function1(seconds):
    time.sleep(seconds)

def кириллица(seconds):
    time.sleep(seconds)

if __name__ == "__main__":
    function1(10)
    кириллица(10)
```

Outputs (`profile.svg` will be created):

```sh
$ py-spy record -o profile.svg -- python3 tests/scripts/cyrillic.py 
py-spy> Sampling process 100 times a second. Press Control-C to exit.


py-spy> Stopped sampling because process exited
py-spy> Wrote flamegraph data to 'profile.svg'. Samples: 12 Errors: 1025
```
</details>

### DIFF
--- a/src/python_data_access.rs
+++ b/src/python_data_access.rs
@@ -33,10 +33,11 @@ pub fn copy_string<T: StringObject, P: ProcessMemory>(
             Ok(chars.iter().collect())
         }
         (2, _) => {
-            // UCS2 strings aren't used internally after v3.3: https://www.python.org/dev/peps/pep-0393/
-            // TODO: however with python 2.7 they could be added with --enable-unicode=ucs2 configure flag.
-            //            or with python 3.2 --with-wide-unicode=ucs2
-            Err(format_err!("ucs2 strings aren't supported yet!"))
+            #[allow(clippy::cast_ptr_alignment)]
+            let chars = unsafe {
+                std::slice::from_raw_parts(bytes.as_ptr() as *const u16, bytes.len() / 2)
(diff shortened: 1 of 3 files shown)


benfred (approved):
thanks for the change!

PR #666, opened 2024-05-16
fix line number for python 3.12

penguin-wwy:
The linetable for 3.12 and 3.11 have the same structure, refactored to a common macro.

Current py-spy dump output on the tests/scripts/longsleep.py test case:
```
Process 1721748: python tests/scripts/longsleep.py
Python v3.12.1 (/data/home/self/.pyenv/versions/3.12.1/bin/python3.12)

Thread 1721748 (idle)
    longsleep (longsleep.py:5)
    <module> (longsleep.py:9)
    <interpreter trampoline> (<shim>:1)
```

### DIFF
--- a/src/python_interpreters.rs
+++ b/src/python_interpreters.rs
@@ -244,6 +244,109 @@ macro_rules! PythonCodeObjectImpl {
     };
 }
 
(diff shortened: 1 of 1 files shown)


benfred (approved):
thanks for this! changes look good to me

PR #672, opened 2024-07-13
(win, python >= 3.10) Fix symbol address resolution

akhramov:
For PE images we use file offsets to calculate symbol addresses in memory. However PE loader uses RVAs (relative virtual addresses) instead, so the current approach breaks for newer python releases.

This change tweaks windows symbol address resolution to rely on RVAs instead of file offsets.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,8 +35,9 @@ serde_derive = "1.0"
 serde_json = "1.0"
 rand = "0.8"
 rand_distr = "0.4"
-remoteprocess = {version="0.4.12", features=["unwind"]}
+remoteprocess = { git = "https://github.com/akhramov/remoteprocess", branch = "feature/windows-aarch64", features = ["unwind"] }
 chrono = "0.4.26"
+cfg-if = "1.0.0"
 
 [dev-dependencies]
 py-spy-testdata = "0.1.0"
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -35,8 +35,9 @@ serde_derive = "1.0"
 serde_json = "1.0"
 rand = "0.8"
 rand_distr = "0.4"
-remoteprocess = {version="0.4.12", features=["unwind"]}
+remoteprocess = { git = "https://github.com/akhramov/remoteprocess", branch = "feature/windows-aarch64" }
```
akhramov, line 38: to-do: Unrelated to this effort. Remove this, needed on my machine only, to be upstreamed in https://github.com/benfred/remoteprocess/pull/89

WoolenWang:
nice try, waiting for this.

benfred:
@akhramov  Thanks so much for the fix!  This PR was very helpful,

This branch has gotten a bit out of sync,  so I've taken the core changes https://github.com/benfred/py-spy/pull/543/commits/ae1e21851e6e8e560c0a674c3e3ed7bff4455500 as part of a PR to add windows CI https://github.com/benfred/py-spy/pull/543 .

PR #679, opened 2024-09-03
Fix clippy warnings

andrewjcg:
```
warning: unexpected `cfg` condition name: `unwind`
  --> src/python_spy.rs:38:11
   |
38 |     #[cfg(unwind)]
   |           ^^^^^^ help: found config with similar value: `panic =
"unwind"`
   |
   = help: consider using a Cargo feature instead
   = help: or consider adding in `Cargo.toml` the `check-cfg` lint
config for the lint:
            [lints.rust]
            unexpected_cfgs = { level = "warn", check-cfg =
['cfg(unwind)'] }
   = help: or consider adding
`println!("cargo::rustc-check-cfg=cfg(unwind)");` to the top of the
`build.rs`
   = note: see
<https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html>
for more information about checking conditional configuration
```

Issue #684, opened 2024-10-09
Error: Failed to find a python interpreter in the .data section


OM152:
* Windows 11
* Python 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)] on win32
* py-spy==0.3.14
* PyCharm 2024.1 (Community Edition) Build #PC-241.14494.241, built on March 28, 2024 Runtime version: 17.0.10+8-b1207.12 amd64

Hi @benfred,
I'm trying to run py-spy.

1. Run my script
```
(.venv) PS W:\PycharmProjects\production_counters_master_2> W:\PycharmProjects\production_counters_master_2\.venv\Scripts\python.exe -m run_prod_counter
```
or 
`(.venv) PS W:\PycharmProjects\production_counters_master_2> python run_prod_counter.py`
Script is running.

2. then I run
`py-spy record -o output.svg --pid 7156`
or
`py-spy top --pid 7156`

3. I get an error
'Error: Failed to find a python interpreter in the .data section'

Can you figure out this problem? Thanks

zanieb:
Is this a duplicate of https://github.com/benfred/py-spy/issues/675

benfred:
This should be fixed by https://github.com/benfred/py-spy/pull/718

benfred:
I think this should be fixed in the 0.4 release. Can you try it out and let me know? 

If this is still not working please re-open the issue

PR #685, opened 2024-10-10
Fix Windows build and lint

zanieb:
@benfred regarding https://github.com/benfred/py-spy/pull/679#issuecomment-2405668108

Includes a few other changes as we debugged the cause of the failure.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,13 +36,13 @@ jobs:
       if: runner.os == 'Linux'
     - name: Build
       run: cargo build --release --verbose --examples
+    - uses: actions/setup-python@v4
+      with:
+        python-version: 3.9
     - name: Test
       id: test
       continue-on-error: true
       run: cargo test --release
-    - uses: actions/setup-python@v4
-      with:
-        python-version: 3.11
     - name: Test (retry#1)
       id: test1
       run: cargo test --release
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1598,7 +1598,7 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf221c93e13a30d793f7645a0e7762c55d169dbb0a49671918a2319d289b10bb"
 dependencies = [
- "windows-sys 0.48.0",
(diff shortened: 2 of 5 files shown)


benfred (approved):
Thanks for the PR!  The changes look good to me,

benfred:
I'd like to get the windows CI working before merging - I realized I might have messed up the how python is installed in the last PR (moved setup-python to the wrong location in the GHA yaml), and am also testing out a tweak to fix in this other PR
  https://github.com/benfred/py-spy/pull/642

zanieb:
Yeah agree. I'm pulling updates from my fork to see if one of those fixes the issue — I don't know what else I would have changed to make CI pass.

zanieb:
It looks like you might have changed from Python 3.9 -> 3.11?

benfred:
I'm sorta wondering if this is related to the python version being tested - I'm getting a different error with python 3.10 than with python 3.11 in CI, and I'm wondering if the working versions are with python 3.9. 

Because of how I messed up the setup python install, its possible that the 'fix clippy' pr we merged tested with python 3.9 (since the install for python 3.11 happened afterwards).

benfred:
> It looks like you might have changed from Python 3.9 -> 3.11?

yeah, I'm testing now on #642 with 3.9 to see. 

The error with 3.10 was `called `Result::unwrap()` on an `Err` value: Failed to find a python interpreter in the .data section
` https://github.com/benfred/py-spy/actions/runs/11279360838/job/31369936001?pr=642 -  which might be the same as https://github.com/benfred/py-spy/issues/684 (so potentially a real issue with python 3.10 on windows, rather than a false positive in CI)

zanieb:
Yeah https://github.com/benfred/py-spy/pull/685/commits/1d0ffc10f760f7d152e757b45f6d7de473345e3b resolves the error.

If you want, I can separate that commit from this pull request. Otherwise, it seems fine to merge this as a whole.

benfred:
I'm good merging this as a whole - thanks for fixing. 

It does like there is an issue on windows with python 3.10+ that needs fixing =( - but  that doesn't need to be fixed in this PR.

PR #688, opened 2024-10-11
Fix bug introduced by commit 5820bf6a7ef56718284add709bc56cc7b8cb6c16

jiangliu:
Commit 5820bf6a7ef56718284add709bc56cc7b8cb6c16 removes field `addr` and `size` from BinaryInfo, which are used when feature `unwind` is enabled. So recover these two fields.

Fix building error:
cargo build --all-features
error[E0609]: no field `addr` on type `&BinaryInfo`
  --> src/binary_parser.rs:18:22
   |
18 |         addr >= self.addr && addr < (self.addr + self.size)
   |                      ^^^^ unknown field
   |
   = note: available fields are: `symbols`, `bss_addr`, `bss_size`

error[E0609]: no field `addr` on type `&BinaryInfo`
  --> src/binary_parser.rs:18:43
   |
18 |         addr >= self.addr && addr < (self.addr + self.size)
   |                                           ^^^^ unknown field
   |
   = note: available fields are: `symbols`, `bss_addr`, `bss_size`

error[E0609]: no field `size` on type `&BinaryInfo`
  --> src/binary_parser.rs:18:55
   |
18 |         addr >= self.addr && addr < (self.addr + self.size)
   |                                                       ^^^^ unknown field
   |
   = note: available fields are: `symbols`, `bss_addr`, `bss_size`

### DIFF
--- a/src/binary_parser.rs
+++ b/src/binary_parser.rs
@@ -10,6 +10,10 @@ pub struct BinaryInfo {
     pub symbols: HashMap<String, u64>,
     pub bss_addr: u64,
     pub bss_size: u64,
+    #[allow(dead_code)]
+    pub addr: u64,
+    #[allow(dead_code)]
+    pub size: u64,
 }
 
 impl BinaryInfo {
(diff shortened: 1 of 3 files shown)


benfred (approved):
thanks for the fix!

I was trying to figure out how this bug got past CI - since we should be building wheels on linux/windows with unwind support for native profiling.

It seems like the native profiling feature has been disabled in the wheels produced by the CI runs with - and trying to use the native profiling feature from the wheels produced by CI will show:

```
$ py-spy record --native
error: Found argument '--native' which wasn't expected, or isn't valid in this context

	If you tried to supply `--native` as a value rather than a flag, use `-- --native`
```

=(.

Anyways, I have a fix for the CI in here https://github.com/benfred/py-spy/pull/691 - 

thanks again for pointing this out

PR #657, opened 2024-03-09
test_coredump: fix unaligned pointer dereference

karolinepauls:
This fixes the test but I still cannot get a coredump to load locally - the executable cannot be located in it, same goes for any other dump I tried. I'll add details in a comment.

The panic observed on master.
```
$ cargo test test_coredump
   Compiling py-spy v0.3.14 (/home/karoline/IDE/py-spy)
    Finished test [unoptimized + debuginfo] target(s) in 4.80s
     Running unittests src/lib.rs (target/debug/deps/py_spy-6453fb3149eb3ee2)

running 1 test
thread 'coredump::test::test_coredump' panicked at src/coredump.rs:94:30:
misaligned pointer dereference: address must be a multiple of 0x8 but is 0x7fa7adbaa334
stack backtrace:
   0: rust_begin_unwind
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_nounwind_fmt::runtime
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:110:18
   2: core::panicking::panic_nounwind_fmt
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:122:9
   3: core::panicking::panic_misaligned_pointer_dereference
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:221:5
   4: py_spy::coredump::CoreDump::new
             at ./src/coredump.rs:94:30
   5: py_spy::coredump::test::test_coredump
             at ./src/coredump.rs:438:20
   6: py_spy::coredump::test::test_coredump::{{closure}}
             at ./src/coredump.rs:434:23
   7: core::ops::function::FnOnce::call_once
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/ops/function.rs:250:5
   8: core::ops::function::FnOnce::call_once
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
thread caused non-unwinding panic. aborting.
error: test failed, to rerun pass `--lib`

Caused by:
[...]

PR #693, opened 2024-10-12
Fix corruption when running cargo test on ARM64

jiangliu:
Run `cargo test` on ARM64 will encounter following corruption, which is caused by unaligned memory access. So enable the test for x86_64 only.

thread 'coredump::test::test_coredump' panicked at src/coredump.rs:94:30: misaligned pointer dereference: address must be a multiple of 0x8 but is 0xffffa6ba0334 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace thread caused non-unwinding panic. aborting.
error: test failed, to rerun pass `--lib`

Caused by:
  process didn't exit successfully: `/dragonball/py-spy.git/target/debug/deps/py_spy-1a1c8047c3d9ee13` (signal: 6, SIGABRT: process abort signal)

### DIFF
--- a/src/coredump.rs
+++ b/src/coredump.rs
@@ -429,12 +429,12 @@ mod elfcore {
     }
 }
 
+#[cfg(target_arch = "x86_64")]
 #[cfg(test)]
 mod test {
     use super::*;
     use py_spy_testdata::get_coredump_path;
 
-    #[cfg(target_pointer_width = "64")]
     #[test]
     fn test_coredump() {
         // we won't have the python binary for the core dump here,


src/coredump.rs
```diff
@@ -429,12 +429,12 @@ mod elfcore {
     }
 }
 
+#[cfg(target_arch = "x86_64")]
 #[cfg(test)]
 mod test {
     use super::*;
     use py_spy_testdata::get_coredump_path;
 
-    #[cfg(target_pointer_width = "64")]
```
zanieb, line 437: Why was this inner cfg insufficient?

benfred:
Does https://github.com/benfred/py-spy/pull/657 fix the issue for you?

src/coredump.rs
jiangliu: ARM64 will also match this `cfg`, but it seems that the coredump file is generated from x86 thus cause unaligned access on ARM64.

jiangliu:
> Does #657 fix the issue for you?

Yeah, missed that PR. I have tried #657 and it does fix the invalid memory access issue. 
BTW, it would be great to generate multiple test coredump files for different archs in crate py-spy-test:)

jiangliu:
Prefer #657 instead of this one.

PR #698, opened 2024-10-15
Format YAML files with Prettier

zanieb:
I was having a hard time reading the Python version changes in https://github.com/benfred/py-spy/pull/696(and more generally, reading the YAML files). If you're interested, here's a format setup with Prettier. I added a pre-commit hook to match the existing lint setup.


benfred (approved):
looks good ! Thanks for also adding the `pre-commit` hook so this stays formatted

Issue #651, opened 2024-01-24
libproc v0.13.0 doesn't build for me


jrmuizel:
```
Caused by:
  process didn't exit successfully: `/Users/jrmuizel/src/py-spy/target/debug/build/libproc-31a6f7d0110b4801/build-script-build` (exit status: 101)
  --- stderr
  /Users/jrmuizel/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libproc-0.13.0/libproc_rs.h:1:10: fatal error: 'libproc.h' file not found
  thread 'main' panicked at 'Failed to build libproc bindings: ClangDiagnostic("/Users/jrmuizel/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libproc-0.13.0/libproc_rs.h:1:10: fatal error: 'libproc.h' file not found\n")', /Users/jrmuizel/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libproc-0.13.0/build.rs:10:10
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  ```
  
  This is fixed by updating to `remoteprocess` 0.4.13

Issue #596, opened 2023-06-26
Failing to build or install py-spy on Mac OS 13


jrabinow:
I'm trying to build py-spy to add it to [MacPorts](https://www.macports.org/), but I seem to be running into an issue with libproc.
Machine info:
```
macOS 13.3 22E252 arm64
Xcode 14.3 14E222b
```

When attempting to `cargo install`, I get the following error:
```
$ cargo install py-spy
error: failed to run custom build command for `libproc v0.12.0`

Caused by:
  process didn't exit successfully: `/var/folders/4q/2c5n_3vx589cd7qdpn31dxfm0000gn/T/cargo-installdn4ixI/release/build/libproc-194877f4dcf5bf8d/build-script-build` (exit status: 101)
  --- stderr
  /Users/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libproc-0.12.0/libproc_rs.h:1:10: fatal error: 'libproc.h' file not found
  /Users/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libproc-0.12.0/libproc_rs.h:1:10: fatal error: 'libproc.h' file not found, err: true
  thread 'main' panicked at 'Failed to build libproc bindings: ()', /Users/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libproc-0.12.0/build.rs:9:10
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
error: failed to run custom build command for `libproc v0.13.0`

Caused by:
  process didn't exit successfully: `/var/folders/4q/2c5n_3vx589cd7qdpn31dxfm0000gn/T/cargo-installdn4ixI/release/build/libproc-0bbbc2f2356e2d3b/build-script-build` (exit status: 101)
  --- stderr
  /Users/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libproc-0.13.0/libproc_rs.h:1:10: fatal error: 'libproc.h' file not found
  thread 'main' panicked at 'Failed to build libproc bindings: ClangDiagnostic("/Users/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libproc-0.13.0/libproc_rs.h:1:10: fatal error: 'libproc.h' file not found\n")', /Users/user/.car
go/registry/src/index.crates.io-6f17d22bba15001f/libproc-0.13.0/build.rs:10:10
[...]

PR #699, opened 2024-10-15
Upgrade `libproc` and `remoteprocess` to fix builds on macOS

zanieb:
Closes https://github.com/benfred/py-spy/issues/596
Closes https://github.com/benfred/py-spy/issues/651

Builds on macOS 13+ are failing because these dependencies are out of date.

I ran:

```
cargo update remoteprocess
cargo update libproc
```

The first was necessary to unblock the `libproc` upgrade. Now, `cargo build` succeeds on my machine.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,6 +161,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "bindgen"
+version = "0.70.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f49d8fed880d473ea71efb9bf597651e77201bdd4893efe54c9e5d65ae04ce6f"
+dependencies = [
+ "bitflags 2.4.1",
+ "cexpr",
+ "clang-sys",
+ "itertools",
+ "proc-macro2",
+ "quote",
+ "regex",
+ "rustc-hash",
+ "shlex",
+ "syn 2.0.38",
+]
+
+[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


benfred (commented):
Thanks for fixing this!

Do we need to upgrade the remoteprocess version in Cargo.toml too ?: https://github.com/benfred/py-spy/blob/6ff51b65cf4a50aeb001fab2e4ce2dd3f41a3922/Cargo.toml#L41

Also can we test this out in CI?  I believe we're only testing macos-13, and I believe switching this to `macos-latest` should do it https://github.com/benfred/py-spy/blob/6ff51b65cf4a50aeb001fab2e4ce2dd3f41a3922/.github/workflows/build.yml#L188

zanieb:
> Do we need to upgrade the remoteprocess version in Cargo.toml too ?:

I don't think we need to. Since it's actually a [version range](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) and the old version is compatible we just want to lock to a newer one during builds.

I can look into a CI test — though I'd loosely prefer to do so separately because I have to make every change on this pull request then rebase onto `master` because otherwise local development is completely broken.

benfred (approved):
>I don't think we need to. Since it's actually a [version range](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) and the old version is compatible we just want to lock to a newer one during builds.

I learned something new today =) .

zanieb:
I don't think adding test coverage there will make a difference — that looks like testing the pre-built wheels whereas this failure occurred while building. We do builds on `macos-latest` in CI so I'm not sure what the difference was. Presumably they are still on an old SDK in the runners or something?

PR #702, opened 2024-10-16
Use new musl cross build Docker images

zanieb:
From https://github.com/benfred/rust-musl-cross/pull/2

Will unblock various dependency upgrades since we won't be stuck on an old toolchain version.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,9 +89,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: [i686-musl, armv7-musleabihf, aarch64-musl, x86_64-musl]
+        target:
+          [
+            i686-unknown-linux-musl,
+            armv7-unknown-linux-musleabihf,
+            aarch64-unknown-linux-musl,
+            x86_64-unknown-linux-musl,
+          ]
     container:
-      image: docker://benfred/rust-musl-cross:${{ matrix.target }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -89,9 +89,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: [i686-musl, armv7-musleabihf, aarch64-musl, x86_64-musl]
+        target:
+          [
+            i686-unknown-linux-musl,
+            armv7-unknown-linux-musleabihf,
+            aarch64-unknown-linux-musl,
+            x86_64-unknown-linux-musl,
```
zanieb, line 97: When published upstream, these tags are shortened. I didn't both with that in the fork.

.github/workflows/build.yml
```diff
@@ -101,15 +107,15 @@ jobs:
       - name: Build
         run: |
           python3 -m pip install --upgrade maturin
-          maturin build --release -o dist --target $RUST_MUSL_CROSS_TARGET --features unwind
+          maturin build --release -o dist --target ${{ matrix.target }} --features unwind
```
zanieb, line 110: Now that the tag matches the target, we can just use it directly which is a little clearer for someone not familiar with the container environment.

benfred (approved):
thanks for this change - looks great!

PR #708, opened 2024-10-19
Python 3.13 Support

benfred:
Initial support for python 3.13 .   This works as far as getting basic stack traces from python 3.13 programs, but there are still some remaining tasks:

- [x] Line numbers are broken 
- [x] Leverage pyruntime.debug_offsets for figuring out interpreter head location 
- [x] Fix cases where PyInterpreterFrame.f_executable doesn't point to a [`PyCodeObject` ](https://github.com/python/cpython/issues/100987#issuecomment-1465963433)
- [x] Thread names 
- [ ] ~~nogil interpreters~~ (#711)


zanieb:
Exciting! Happy to review this when it's ready, if you want.

benfred:
@zanieb - that would be great, I'd appreciate it! I'll ping you when its ready

benfred:
This doesn't fully handle freethreaded/nogil python interpreters - but I'm leaning towards handling that in a separate PR and getting this merged in first (I also suspect that this will mostly work with freethreaded python interpreters, just without thread names coming up , but I haven't tested that yet).

One other issue is that the `f_executable` pointer on the FrameObject can be any arbitrary python object in python3.13 - and not just a PyCodeObject like in older versions of python. This PR is currently just skipping frames where we can't read the filename and function name (indicating a f_executable that isn't a PyCodeObject) -but maybe should instead be checking if the `f_executable->ob_type == &PyCode_Type` like `PyCode_Check` does in cpython (?).  (The downside to checking the type is that we'd have to figure out the address of PyCode_Type in the target process and pass along to the get_stack_traces function - which might be a bit difficult right now)

@zanieb fyi

zanieb:
Cool I'll review this week.

zanieb (approved):
I don't think my review is worth much, but I didn't see anything concerning.

PR #710, opened 2024-10-23
Test macos py3.11

benfred:
It doesn't seem like py3.13 has the same SIP restrictions on OSX in GHA that we've been seeing previously, so try enabling 3.11 & 3.12 in CI


benfred:
The python 3.12 tests are failing on macos on failing to find the python version  - which I can't replicate locally on my mac.  (and both python 3.11 and python 3.13 seem to be working fine in CI).

Issue #709, opened 2024-10-19
Fail to find interpreter with Standalone Python and shapely.geos


jack-zhang-ai:
We are hitting the following issue when trying to py-spy while using a Python distribution from [python-build-standalone](https://gregoryszorc.com/docs/python-build-standalone/main/index.html#) and importing shapely.geos from shapely==1.8.5.

```
$ sudo env "PATH=$PATH" RUST_LOG=info py-spy dump --pid 148674
[2024-10-18T23:25:09.297598685Z INFO  py_spy::config] Command line args: ArgMatches { args: {}, subcommand: Some(SubCommand { id: [hash: B8461C91A07ADDC8], name: "dump", matches: ArgMatches { args: {[hash: CD5160AB4406C427]: MatchedArg { occurs: 1, source: Some(CommandLine), indices: [2], type_id: Some(TypeId { t: 15469221632486072992 }), vals: [[AnyValue { inner: TypeId { t: 15469221632486072992 } }]], raw_vals: [["148674"]], ignore_case: false }}, subcommand: None } }) }
[2024-10-18T23:25:09.297957181Z INFO  py_spy::python_spy] Got virtual memory maps from pid 148674:
[2024-10-18T23:25:09.298431475Z INFO  py_spy::python_spy] Found libpython binary @ /python_standalone_testing/python/lib/libpython3.9.so.1.0
[2024-10-18T23:25:09.311940673Z INFO  py_spy::python_spy] got symbol Py_GetVersion.version (0x00007f23d9445860) from libpython binary
[2024-10-18T23:25:09.311949009Z INFO  py_spy::python_spy] Getting version from symbol address
[2024-10-18T23:25:09.312385612Z INFO  py_spy::python_spy] Getting version from python binary BSS
[2024-10-18T23:25:09.312401592Z INFO  py_spy::python_spy] Failed to get version from BSS section: failed to find version string
[2024-10-18T23:25:09.312406531Z INFO  py_spy::python_spy] Getting version from libpython BSS
[2024-10-18T23:25:09.312561654Z INFO  py_spy::python_spy] Failed to get version from libpython BSS section: failed to find version string
[2024-10-18T23:25:09.312578105Z INFO  py_spy::python_spy] Trying to get version from path: /python_standalone_testing/python/bin/python3.9
[2024-10-18T23:25:09.312583014Z INFO  py_spy::python_spy] python version 3.9.0 detected
[...]

PR #715, opened 2024-10-30
Test BSS scanning code

benfred:
Seems like we can't figure out the python interpreter address w/o symbols, which definitely used to work . Testing this out in CI to figure out what version this started to fail at


benfred:
Seems like the BSS scanning breaks with python 3.10+ and works for older versions of python. 

This means that we currently require some debug symbols for python 3.10+, which is might be the root cause of issues we've been seeing on windows w/ python 3.10+ (and maybe even things like https://github.com/benfred/py-spy/issues/709 ).

benfred:
running git bisect on my linux box using this branch - shows that this change broke the BSS scanning feature https://github.com/python/cpython/pull/4802 , by moving the PyRuntimeState from the BSS section to its own named section

Issue #593, opened 2023-06-15
"Failed to find a python interpreter in the .data section"


xiewendan:
I use py-spy to run py by python3.11.3 on windows

`
py-spy record -o profile.svg -- python -c "import time;time.sleep(100)"
Error: Failed to find a python interpreter in the .data section
`

operating system: windows
py-spy version: 0.3.14
python version: 3.11.3

akjroller:
I think the issue is that py-spy only supports Python versions up to 3.9. I tried it as well using 3.11.4 and got the same error.

alispa:
Thanks @akjroller for at least the temporary solution.
I confirm that on python 3.9 it worked and on python 3.10 not. (Windows)

burrowsa:
I am also seeing this issue. If the program I want to profile is running in python 3.10 then I can't profile it with py-spy, but I can profile the same program running under python 3.7

![image](https://github.com/benfred/py-spy/assets/1609099/e226d167-3312-4c27-be13-94628215f8bd)
![image](https://github.com/benfred/py-spy/assets/1609099/c298ebe7-cea7-4d14-9680-e1555cbc3b7a)

aakash-sahai:
This is a duplicate of https://github.com/benfred/py-spy/issues/643. Some solutions/workarounds can be found posted there.

fleimgruber:
@aakash-sahai thanks for mentioning! But measured by creation time, #643 is a duplicate of this issue, also mentioned in https://github.com/benfred/py-spy/issues/643#issuecomment-1859271016.

akhramov:
https://github.com/benfred/py-spy/pull/672 this __might__ fix the problem. I need to test on an amd64 machine first.

benfred:
I think this should be fixed in the 0.4 release. Can you try it out and let me know?

If this is still not working please re-open the issue

PR #718, opened 2024-10-31
Fix profiling when python symbols aren't available.

benfred:
Since python 3.10 - we haven't been able to profile python interpreters that have been compiled without symbols. This is because cpython changed where the 'PyRuntime' global is stored in python 3.10, from being in the BSS section into being in its own named section in the binary (https://github.com/python/cpython/pull/4802)

This especially affected profiling on windows, where you'd have to install python symbols to be able to use py-spy.

Fix by reading in the address/size of the the PyRuntime section from the elf/mach/pe binaries and using that to scan python interpreters when symbols aren't available.

### DIFF
--- a/src/binary_parser.rs
+++ b/src/binary_parser.rs
@@ -12,6 +12,8 @@ pub struct BinaryInfo {
     pub symbols: HashMap<String, u64>,
     pub bss_addr: u64,
     pub bss_size: u64,
+    pub pyruntime_addr: u64,
+    pub pyruntime_size: u64,
     #[allow(dead_code)]
     pub addr: u64,
     #[allow(dead_code)]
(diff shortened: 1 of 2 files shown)


zanieb:
I believe you can also test this with the "stripped" variants of `python-build-standalone` (which uv installs by default)

benfred:
it would be pretty cool if we could test this out using a stripped build in CI - does the uv build strip out the PyRuntime symbol too ? (like what does something like `nm -A `which python` | grep PyRuntime` show ?)

In the short term, I'm testing this by verifying this fixes the windows wheels https://github.com/benfred/py-spy/pull/543 - since this mostly impacts windows right now

benfred:
I've created https://github.com/benfred/py-spy/issues/719 to track testing this functionality in CI - but I'm going to merge this change in now since it does fix the issue , and I've tested this manually to verify it works (and I'd like to get a release out with =)

fleimgruber:
@benfred Just a quick heads-up on https://github.com/benfred/py-spy/pull/718#issuecomment-2451054736:

I am still seeing #593.
I tried with a Python installed by uv and got:

```
/usr/bin/nm: /c/Users/LeimgruberF/lg/.venv/Scripts/python: no symbols
```

This is on Windows (using `nm` from MSYS2 MinGW64).

PR #751, opened 2025-03-15
Fix native stack merging for Python 3.12/3.13

krfricke:
Python changed the way how entry frames are detected yet again in Python 3.12+, see https://github.com/python/cpython/issues/96421.

Now, shim frames are inserted before every block of Python frames instead of marking the first frame as `is_entry`.

Because `is_entry` is now effectively unused, merging multiple blocks of native and python frames didn't work anymore.

To resolve, we add a new field to the `Frame` struct called `is_shim_entry`, which is set when the frame was preceded by a shim frame. Because we iterate in reverse, we have to set this for the last element in the frames vector.

We could also re-use the `is_entry` field. I went with a new field because `is_entry` is still technically part of the Python frame spec and may be re-used in the future (though I think this is unlikely).


krfricke:
I can look into adding a test for the nested native - python merge logic, but have to see if I can find the bandwidth to do this properly.

benfred (approved):
lgtm - thanks for the fix!

Issue #735, opened 2024-12-10
panic with Python 3.12 on Linux


bouweandela:
When trying to profile my Python program with py-spy 0.4.0, the tool crashes with the following stack trace:

```
thread '<unnamed>' panicked at src/python_interpreters.rs:257:16:
index out of bounds: the len is 11 but the index is 11
stack backtrace:
   0:     0x7ffff7c28075 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h358afad87e02ca76
   1:     0x7ffff7c6277b - core::fmt::write::hb19b5b269a2fe458
   2:     0x7ffff7c2545f - std::io::Write::write_fmt::he5a92676a45ef09d
   3:     0x7ffff7c291c1 - std::panicking::default_hook::{{closure}}::h3bff550b24d93725
   4:     0x7ffff7c28e9c - std::panicking::default_hook::hd53b1b06d2b99687
   5:     0x7ffff7c29821 - std::panicking::rust_panic_with_hook::h9fdd87cddb2763da
   6:     0x7ffff7c29687 - std::panicking::begin_panic_handler::{{closure}}::h089783ab6b5cba45
   7:     0x7ffff7c28539 - std::sys::backtrace::__rust_end_short_backtrace::hed34776d77ef7922
   8:     0x7ffff7c29314 - rust_begin_unwind
   9:     0x7ffff7872453 - core::panicking::panic_fmt::h300583f35f37447a
  10:     0x7ffff78725e7 - core::panicking::panic_bounds_check::hbdc1cc1e8fde102f
  11:     0x7ffff796ee9e - <py_spy::python_bindings::v3_12_0::PyCodeObject as py_spy::python_interpreters::CodeObject>::get_line_number::h365860ffb62d2018
  12:     0x7ffff78c750a - py_spy::stack_trace::get_stack_trace::hf218688a270a5d72
  13:     0x7ffff790a26c - py_spy::python_spy::PythonSpy::get_stack_traces::he097640af52b1cd3
  14:     0x7ffff78cb9bc - std::sys::backtrace::__rust_begin_short_backtrace::hb23a7328ba172b05
  15:     0x7ffff78cd11b - core::ops::function::FnOnce::call_once{{vtable.shim}}::h0aa6e250e2ff7269
  16:     0x7ffff7c2dedb - std::sys::pal::unix::thread::Thread::new::thread_start::h4233e449bef376b8
[...]

Issue #569, opened 2023-04-23
panic with top


shaolo1:
thread '<unnamed>' panicked at 'index out of bounds: the len is 77 but the index is 77', src/python_interpreters.rs:296:20
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Any { .. }', src/sampler.rs:218:22

invoked via:
sudo /home/foov/venv311/bin/py-spy top --nonblocking --pid 417984

Jongy:
Can you run with `RUST_BACKTACE=1`, i.e `sudo RUST_BACKTACE=1 /home/foov/venv311/bin/py-spy top --nonblocking --pid 417984` and post the full output?

shaolo1:
Where does the backtrace go?  I don't see any change to the output.

shaolo1:
...with RUST_BACKTRACE=full

thread '<unnamed>' panicked at 'index out of bounds: the len is 3 but the index is 3', src/python_interpreters.rs:296:20
stack backtrace:
   0:     0x7fea05c01dfd - std::backtrace_rs::backtrace::libunwind::trace::hc5376e1f9e4c14fa
                               at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x7fea05c01dfd - std::backtrace_rs::backtrace::trace_unsynchronized::hd15c3fc9f4947abd
                               at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x7fea05c01dfd - std::sys_common::backtrace::_print_fmt::ha8075890ce48b75c
                               at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/std/src/sys_common/backtrace.rs:66:5
   3:     0x7fea05c01dfd - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hc0b18c49c3f409e9
                               at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/std/src/sys_common/backtrace.rs:45:22
   4:     0x7fea05c3d82c - core::fmt::write::h5f97ce7fceb49c8e
[...]

PR #755, opened 2025-04-01
Formalize linetable parsing and propagate out-of-bounds errors

noahbkim:
We've noticed a transient issue where `py-spy` will abort due to unmet assumptions about the contents of `co_linetable` (for example we've seen `[231]`, which doesn't even satisfy the internal specification). To address this, I've patched the `co_linetable` parsing to propagate out-of-bounds errors so that the samples are simply ignored (and a warning is raised).

This change is not as minimal as it could be, for example it reorders a couple lines in `python_interpreters.rs` for clarity. If that's a problem, I'd be happy to go back and pare down such changes.

Taking a glance, I would guess this is the root cause for https://github.com/benfred/py-spy/issues/735 and https://github.com/benfred/py-spy/issues/569, although I have no explanation for why the `co_linetable`'s appear invalid.


src/python_interpreters.rs
```diff
@@ -65,7 +67,7 @@ pub trait CodeObject {
     fn argcount(&self) -> i32;
     fn varnames(&self) -> *mut Self::TupleObject;
 
-    fn get_line_number(&self, lasti: i32, table: &[u8]) -> i32;
+    fn get_line_number(&self, lasti: i32, table: &[u8]) -> Result<i32, Error>;
```
Hnasar, line 70: is there a reason not to just do this simpler form everywhere? 
```suggestion
    fn get_line_number(&self, lasti: i32, table: &[u8]) -> Result<i32>;
```

https://docs.rs/anyhow/latest/anyhow/#details

src/python_interpreters.rs
```diff
@@ -240,34 +242,89 @@ macro_rules! PythonCodeObjectImpl {
...
+        }
+    }
+
+    fn try_read_header(&mut self) -> Result<Option<u8>, Error> {
+        match self.try_read() {
+            Some(byte) => {
+                if byte & 0b10000000 == 0 {
+                    Err(Error::msg(format!(
+                        "Expected header bit at cursor {} in {:?}",
+                        self.cursor, self.buffer
```
Hnasar, line 287:
```suggestion
                        self.cursor-1, self.buffer
```

in `try_read` you advance the `cursor`, but if there's a wrong byte, you want to report on the one you just read, I think.

src/python_interpreters.rs
```diff
@@ -240,34 +242,89 @@ macro_rules! PythonCodeObjectImpl {
...
+                    Ok(Some(byte))
+                }
+            }
+            None => Ok(None),
+        }
+    }
+
+    fn read_body(&mut self) -> Result<u8, Error> {
+        let byte = self.read()?;
+        if byte & 0b10000000 == 1 {
```
Hnasar, line 299: this will never be true
```suggestion
        if byte & 0b10000000 != 0 {
```

korniltsev:
> This change is not as minimal as it could be, for example it reorders a couple lines in python_interpreters.rs for clarity. If that's a problem, I'd be happy to go back and pare down such changes.

Let's change only what's needed for the fix, so that it's easier to cherry-pick & revert & reason about

src/python_interpreters.rs
noahbkim: it's written out like this throughout the rest of the codebase, better to stay consistent

Hnasar:
@benfred can approve this workflow so the build can run?

Otherwise this LGTM, and we've manually built it and confirmed it's fixed the issues we were seeing.

korniltsev:
I suggest we also fix int overflows either in the same PR or in a followup, so that debug builds of pyspy will not panic

korniltsev:
>  @korniltsev can you point to the overflows to fix?

Basically every time some uncontrolled user-provided number (from the linetable) is added, subtracted, or shifted, etc.
You can search for "checked" in this diff https://github.com/grafana/py-spy/pull/1/files
Fuzzing is a great and easy way to find some of those overflows.

Btw, I am not a maintainer here, this is just a suggestion for improvements. Maybe even better to do as a separate change. Feel free to ignore.

PR #758, opened 2025-04-07
fix: alow library users to disable remoteprocess/unwind feature

korniltsev:
This should allow to use pyspy as a library without linking into libunwind like this
```
py-spy = { version = "0.4.0", default-features = false }
```


korniltsev:
I am not alone https://github.com/benfred/py-spy/pull/768/files

Issue #756, opened 2025-04-02
Failed to find python version from target process (py-spy 0.3.14, python 3.11)


le-cardinal:
hi,

getting a cmmon error running my container based on a alpine-python linux image, so installation is done with `RUN apk add py-spy --update-cache --repository http://dl-3.alpinelinux.org/alpine/edge/testing/ --allow-untrusted`

After trying to enter into the container and using py-spy, i got an error:
'py-spy top --pid 10
Error: Failed to find python version from target process'

Any idea ?

Regards
/daniel

bittner:
I have the same problem with [uv-installed Pythons](https://docs.astral.sh/uv/guides/install-python/), e.g.

```console
$ uv run py-spy top -- uv run matrix-validator tests/fixtures/valid-3x2.csv

Error: Failed to find python version from target process
```

```console
$ uv run py-spy record -o profile.svg -- uv run matrix-validator tests/fixtures/valid-3x2.csv
Error: Failed to find python version from target process
$ ...  # (my CLI application's output follows)
```

sander76:
In this case I activate the env manually and run pyspy without any mention of uv. Seems to work fine.

eduardz1:
> I have the same problem with [uv-installed Pythons](https://docs.astral.sh/uv/guides/install-python/), e.g.
> 
> $ uv run py-spy top -- uv run matrix-validator tests/fixtures/valid-3x2.csv
> 
> Error: Failed to find python version from target process
> 
> $ uv run py-spy record -o profile.svg -- uv run matrix-validator tests/fixtures/valid-3x2.csv
> Error: Failed to find python version from target process
> $ ...  # (my CLI application's output follows)

have you managed to find a workaround?

stasmalinowski:
With uv one thing I find helpful is adding the `-s` flag. It seems that when you point py-spy at `uv run matrix-validator ...` py-spy looks at the uv process instead of the python process that uv spawns. When you specify `-s`, py-spy will also profile subprocesses, so it finds the python process spawned by uv.

Issue #550, opened 2023-02-04
Error: Failed to find python version from target process


kDaeMN:
Getting below stack when trying to use py-spy on

DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.1 LTS"

Python 2.7.18

py-spy 0.3.14

 INFO 2023-02-04T14:36:42Z: py_spy::python_spy: Found libpython binary @ /opt/XX/YYY/lib/libpython2.7.so.1.0
 INFO 2023-02-04T14:36:42Z: py_spy::python_spy: Getting version from python binary BSS
 INFO 2023-02-04T14:36:42Z: py_spy::python_spy: Failed to get version from BSS section: failed to find version string
 INFO 2023-02-04T14:36:42Z: py_spy::python_spy: Getting version from libpython BSS
 INFO 2023-02-04T14:36:42Z: py_spy::python_spy: Failed to get version from libpython BSS section: failed to find version string
 INFO 2023-02-04T14:36:42Z: py_spy::python_spy: Trying to get version from path: /opt/XX/YYY/lib/EXECUTABLENAME
Error: Failed to find python version from target process

gobater:
There is a similar bug: https://github.com/benfred/py-spy/issues/550

This problem also occurs on windows when calling a module entry point (executable) instead of directly calling the python interpreter

itamarst:
I am seeing this on a bunch of module entry point executables on Python 3.10 on Windows, as above. Austin does manage to do it, so I assume it's possible.

ses4j:
Be aware `--subprocesses` may help if you're using a venv, but doesn't help with 3.10+

fleimgruber:
@gobater I assume you meant to link to this similar bug: https://github.com/benfred/py-spy/issues/498?

gobater:
> @gobater I assume you meant to link to this similar bug: #498?

Yes, you are correct. I mixed it up while c&p the link

PR #765, opened 2025-05-09
[Fix] Symbol mapping issue when we have multiple executable segment

yinghai:
It turns out that we can have multiple executable segment for python binaries
```
➜  py-spy git:(master) ✗ grep libpython3 /proc/3033958/maps
7b9d5718a000-7b9d5718b000 r-xp 0065e000 09:00 271849319                  /tmp/yinghai/cpython-3.12.9-linux-x86_64-gnu/lib/libpython3.12.so.1.0
7b9d580f1000-7b9d58311000 r--p 00000000 09:00 271849319                  /tmp/yinghai/cpython-3.12.9-linux-x86_64-gnu/lib/libpython3.12.so.1.0
7b9d58311000-7b9d5902e000 r-xp 00220000 09:00 271849319                  /tmp/yinghai/cpython-3.12.9-linux-x86_64-gnu/lib/libpython3.12.so.1.0
7b9d5902e000-7b9d5962a000 r--p 00f3d000 09:00 271849319                  /tmp/yinghai/cpython-3.12.9-linux-x86_64-gnu/lib/libpython3.12.so.1.0
7b9d5962a000-7b9d5974f000 r--p 01538000 09:00 271849319                  /tmp/yinghai/cpython-3.12.9-linux-x86_64-gnu/lib/libpython3.12.so.1.0
7b9d5974f000-7b9d598e6000 rw-p 0165d000 09:00 271849319                  /tmp/yinghai/cpython-3.12.9-linux-x86_64-gnu/lib/libpython3.12.so.1.0
```
Note that both segement `7b9d5718a000` and `7b9d58311000` are executable but first segment has a lower address but a higher offset than the second one. In the code to parse binary, we find the first executable with lowerest address but in symbol map resolution, we use elf in linux to find the first executable with lowest offset to get absolute address of a symbol.
https://github.com/benfred/py-spy/blob/1fa3a6ded252d7c1c0ff974a4fcd1af67a1577cf/src/binary_parser.rs#L139-L145

This inconsistency causes the symbol address to wrong. Hence py-spy will fail to find python context of the process. Typical error message is 
> Failed to find python version from target process

We have a few of such issues (https://github.com/benfred/py-spy/issues/756, https://github.com/benfred/py-spy/issues/550), which could be related although there are other reasons that can lead to this. 

The fix here is to scan all the executable segments and pick the lowest offset one to parse binary so that we are consistent.

### DIFF
--- a/src/python_process_info.rs
+++ b/src/python_process_info.rs
@@ -143,24 +143,27 @@ impl PythonProcessInfo {
 
         // likewise handle libpython for python versions compiled with --enabled-shared
         let libpython_binary = {
-            let libmap = maps.iter().find(|m| {
-                if let Some(pathname) = m.filename() {
-                    if let Some(pathname) = pathname.to_str() {
-                        #[cfg(not(windows))]
-                        {
-                            return is_python_lib(pathname) && m.is_exec();
-                        }
-                        #[cfg(windows)]
-                        {
-                            return is_python_lib(pathname);
+            let libmaps: Vec<_> = maps
+                .iter()
+                .filter(|m| {
+                    if let Some(pathname) = m.filename() {
+                        if let Some(pathname) = pathname.to_str() {
+                            #[cfg(not(windows))]
+                            {
+                                return is_python_lib(pathname) && m.is_exec();
+                            }
+                            #[cfg(windows)]
+                            {
+                                return is_python_lib(pathname);
+                            }
                         }
                     }
-                }
-                false
-            });
+                    false
+                })
+                .collect();
 
             let mut libpython_binary: Option<BinaryInfo> = None;
-            if let Some(libpython) = libmap {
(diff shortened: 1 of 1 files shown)


korniltsev:
Can you please share the libpython binary or a way to reproduce / obtain it?

korniltsev:
Nit: It would be nice to have a regression test.

yinghai:
I actually don't think this is library specific. It tends to happen on the spawned subprocess. It's just the binaries go remapped and the segments are reordered in terms of memory address vs offset.

yinghai:
Pushed a change to keep the behavior unchanged for windows.

fleimgruber:
@yinghai thanks for the quick change, but I still get "Error: Failed to find python version from target process" on Python 3.10.11

fleimgruber:
@benfred The addressed issues are still unresolved for Windows - should we create a tracking issue for a similar fix for Windows?

PR #769, opened 2025-06-25
docs: use debug ephemeral containers

duhow:
Explains to use **ephemeral containers** to attach to a running Pod with permissions, so that way you can run `py-spy` live on the running application without needing to restart it.


benfred:
thanks!

Issue #753, opened 2025-03-26
No stack counts found on ARM Python 3.12


HiFiPhile:
Hello,

I'm trying to run py-spy on embedded Linux built by Yocto 5.0 with SAM9X60 CPU (ARM926).

- glibc 2.39
- Python 3.12.9
- rustc 1.75

Also tried #751 applied.

Flamechart returns `No stack counts found`
```
py-spy> Stopped sampling because Control-C pressed
[2025-03-26T11:03:40.370604320Z ERROR inferno::flamegraph] No stack counts found
```

Dumps returns all threads as idle:
```
/py-spy dump --pid 17694
Process 17694: python /hello.py
Python v3.12.9 (/usr/bin/python3.12)

Thread 17694 (idle)
Thread 17699 (idle)
Thread 17700 (idle)
Thread 17704 (idle)
Thread 17705 (idle)
```

skrap:
Looks like some problem in the stack trace walking.  I added 
```
+        debug!("filename: {:?}", &filename);
+        debug!("name: {:?}", &name);
        if filename.is_err() || name.is_err() {
            frame_ptr = frame.back();
            continue;
        }
+        debug!("found filename and name");
```
in `get_stack_trace<T, P>()` and on the armv7-unknown-linux-gnueabihf target, with python3.12, I *never* get past that `continue`. 

The errors all look like bad memory accesses:

```[2025-06-23T18:35:23.126544219Z DEBUG py_spy::stack_trace] filename: Err(Failed to copy filename
    
    Caused by:
        Refusing to copy 787480704 chars of a string)
[2025-06-23T18:35:23.126948552Z DEBUG py_spy::stack_trace] name: Err(Failed to copy function name
    
    Caused by:
        0: Bad address (os error 14)
        1: Bad address (os error 14))
```

I'm looking at the workarounds around extra levels of indirection in the stack frame with some suspicion but I'm very ignorant about cpython internals, so if someone has some insight, I'd be glad for it!

skrap:
I found the issue.  This is a problem with the generated `src/python_bindings/v3_12_0.rs`.

In python, the `struct _object` has an extra field `ob_refcnt_split` on 64-bit systems.  The bindgen output doesn't take this into account.

See https://github.com/python/cpython/blob/3.12/Include/object.h#L182 for the python definition.

As a test, changing the definition of `_object__bindgen_ty_1` to be the following allows py-spy to run correctly:
```
pub union _object__bindgen_ty_1 {
[...]

PR #773, opened 2025-07-11
fix stack trace fetching on 32-bit cpython 3.12 and 3.13

skrap:
The stack traces fail because 32-bit and 64-bit python define struct _object differently.

See https://github.com/benfred/py-spy/issues/753#issuecomment-3006045678 for the root cause.

Fixes https://github.com/benfred/py-spy/issues/753

This is not robust against regeneration of the python bindings, but will fix things on 32-bit for now.

### DIFF
--- a/src/python_bindings/v3_12_0.rs
+++ b/src/python_bindings/v3_12_0.rs
@@ -391,9 +391,11 @@ pub struct _object {
 #[derive(Copy, Clone)]
 pub union _object__bindgen_ty_1 {
     pub ob_refcnt: Py_ssize_t,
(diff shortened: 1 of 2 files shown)


benfred:
I'm good with this change - but am unsure the best way to test it out (like so that we don't just have the same issue when we add python 3.14 support etc).

I've added aarch64 support in CI recently here https://github.com/benfred/py-spy/pull/775 - but I don't see GHA runners available with 32-bit arm support in them =(

I could run a self hosted armv7 runner again for testing , and I previously hosted this on a raspberry pi I own - but this was brittle and broke all the time (and didn't test all the python combinations, so wouldn't have caught this anyways).

We could also probably use something like qemu to test this in a virtualized setting on GHA (?).

PR #783, opened 2025-07-21
RFC OTLP collector continuous export

korniltsev:
- Adds `--continuous` flag
- Adds otlp Recorder implementation https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/collector/profiles/v1development/profiles_service.proto  depends on https://github.com/open-telemetry/opentelemetry-rust/pull/3077

This is very much early draft, not well tested, etc. Would like to have some early feedback if this is something could be considered merged.


korniltsev-grafanista:
@benfred may we have some feedback please? Is this something that will be considered for merge? Should we continue and finish this draft or close the PR?

Now that otel profiles are in alpha state they are less likely to break https://opentelemetry.io/blog/2026/profiles-alpha/

PR #786, opened 2025-07-23
Allow specifying the `--pid` commandline option as hex

benfred:


src/config.rs
```diff
@@ -433,9 +433,15 @@ impl Config {
         config.command = subcommand.to_owned();
 
         // options that can be shared between subcommands
-        config.pid = matches
-            .value_of("pid")
-            .map(|p| p.parse().expect("invalid pid"));
+        config.pid = matches.value_of("pid").map(|p| {
+            // allow pid to be specified as a hexadecimal value
+            if p.to_lowercase().starts_with("0x") {
+                Pid::from_str_radix(&p[2..], 16).expect("invalid pid")
```
zanieb, line 439: You could use `strip_prefix` instead of slicing after the `starts_with`.

PR #788, opened 2025-07-27
fix test_negative_linenumber_increment assertion for py>=3.12

karolinepauls:
Comprehensions got inlined - https://peps.python.org/pep-0709/


benfred:
thanks for the fix! 

fwiw, we're not running these rust integration tests on every version of python in CI (just the python integration tests, which test out the py-spy CLI installed from the generated wheels) - which is why this wasn't detected

Issue #785, opened 2025-07-21
Stack overflow when profiling Python 3.12.3


Timothy-Liuxf:
I run a python program with Python 3.12.3:

<img width="475" height="385" alt="Image" src="https://github.com/user-attachments/assets/c7a31c94-68f6-471b-bd37-0677fee37e67" />

And I profile it with py-spy (commit hash: 405b56a513e169fc0fb45ec42f610ce8a7eee53e):

```bash
$ sudo ./py-spy record profile.svg --pid 136773
```

And an stack overflow error occurred:

```

thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
Aborted
```

<img width="605" height="121" alt="Image" src="https://github.com/user-attachments/assets/48913b13-90f7-419d-8653-0cadad9fa845" />

However, when I run it with Python 3.8.12, it works:

<img width="502" height="241" alt="Image" src="https://github.com/user-attachments/assets/2f310404-8a44-4ab7-bc35-c8455fa6992f" />

<img width="907" height="95" alt="Image" src="https://github.com/user-attachments/assets/70a9c691-424a-4289-97d6-eaf7b5262459" />

karolinepauls:
Since stack overflow happens in a thread, setting RUST_MIN_STACK higher than 2MB (in bytes) at runtime is a workaround.

e.g. 
```
RUST_MIN_STACK="$(( 4 * 1024 * 1024 ))" ./target/debug/py-spy record python3.12 ...
```
No workaround needed for release builds.

PR #789, opened 2025-07-27
[draft] don't copy all of PyInterpreterState

karolinepauls:
This fixes [785](https://github.com/benfred/py-spy/issues/785) - stack overflow with Python 3.12 observed in debug builds.

Similarly to the reporter, I saw it locally on amd64 Linux, debug build. *It doesn't happen in release builds*.

## Analysis

It appears that the error happens when copying PyInterpreterState which is larger in 3.12 compared to the other pythons:
```
std::mem::size_of::<v3_10_0::_is>() = 113520
std::mem::size_of::<v3_11_0::_is>() = 107736
std::mem::size_of::<v3_12_0::_is>() = 383520
std::mem::size_of::<v3_13_0::_is>() = 194968
```
it's 374.53125 kiB.

Additionally, Rust is known for bloated stack usage in debug builds due to unoptimised pass-by-copy (in my understanding at least) and issues like this https://github.com/rust-lang/rust/issues/34283.

The stack sizes I observed in gdb are about 2MB. This corresponds to the default stack size in nptl (the stack overflow happens in thread 2, not the main thread).

**Stack trace with frame pointers:**
```
frame                                                                                                   stack_pointer   diff
../sysdeps/posix/raise.c:26                                                                             0x7b9d4d4ba2b0  -32
./stdlib/abort.c:73                                                                                     0x7b9d4d4ba2d0  -208
library/std/src/sys/pal/unix/mod.rs:366                                                                 0x7b9d4d4ba3a0  -16
library/std/src/rt.rs:50                                                                                0x7b9d4d4ba3b0  -208
<signal_handler_called>                                                                                 0x7b9d4d4ba480  
/home/karoline/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/remoteprocess-0.5.0/src/lib.rs:188  0x7b9d4cf08878  -348168
src/python_process_info.rs:618                                                                          0x7b9d4cf5d880  -1736096
src/python_spy.rs:65                                                                                    0x7b9d4d105620  -2784
src/python_spy.rs:107                                                                                   0x7b9d4d106100  -1584
src/sampler.rs:51                                                                                       0x7b9d4d106730  -2576
/rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/sys/backtrace.rs:152                    0x7b9d4d107140  -16
/rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/thread/mod.rs:559                       0x7b9d4d107150  -288
/rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panic/unwind_safe.rs:272               0x7b9d4d107270  -256
/rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:589                        0x7b9d4d107370  -272
__rust_try                                                                                              0x7b9d4d107480  -32
/rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:552                        0x7b9d4d1074a0  0
/rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panic.rs:359                            0x7b9d4d1074a0  0
/rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/thread/mod.rs:557                       0x7b9d4d1074a0  -1984
/rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/ops/function.rs:250                    0x7b9d4d107c60  -32
library/alloc/src/boxed.rs:1966                                                                         0x7b9d4d107c80  0
library/alloc/src/boxed.rs:1966                                                                         0x7b9d4d107c80  0
library/std/src/sys/pal/unix/thread.rs:97                                                               0x7b9d4d107c80  -80
./nptl/pthread_create.c:448                                                                             0x7b9d4d107cd0  -176
../sysdeps/unix/sysv/linux/x86_64/clone3.S:78                                                           0x7b9d4d107d80  
```
This is 2094344B (2045.2578 kiB) in total (from the beginning to stack to the `copy_struct` function in remoteprocess which overflowed) - your numbers may differ a little because I placed 2 1024-byte long bytearrays on the stack in `pyspy_main` in order to have addresses to pass to `pthread_attr_getstacksize` in the next point.

**Double-checking thread size at runtime:**
```
(gdb) set backtrace past-main
... `up` enough times to get to nptl, then write to stack addresses reserved for this step
(gdb) p pthread_attr_getstacksize(0x7fffffffb610, 0x7fffffffb210)
$3 = 0
(gdb) p *(size_t*)0x7fffffffb210
$4 = 2048000
```

## Solution

I decided to, instead of copying `PyInterpreterState`, calculate offsets of struct of interest and copy only these structs. The error stopped manifesting after removing copying from `get_threadstate_address` and `PythonSpy::_get_stack_traces` but I removed all copying of `PyInterpreterState`.  I didn't make it very pretty by refactoring anything - `InterpreterState` now has `threadstate_ptr_ptr` and `modules_ptr_ptr` methods returning remote pointers to struct entries (which are both double pointers).

This should improve profiling performance somewhat but most likely it's the number of copies that matters, rather the size of a copy.

## Alternatives

1. Setting a higher stack size could mask this issue for some time.
2. Heap allocation would solve the issue, but require a similar number of changes.

## Questions

Is there something I may have broken that tests won't see?

### DIFF
--- a/src/coredump.rs
+++ b/src/coredump.rs
@@ -326,11 +326,18 @@ impl PythonCoreDump {
     }
 
     fn _get_stack<I: InterpreterState>(&self, config: &Config) -> Result<Vec<StackTrace>, Error> {
-        let interp: I = self.core.copy_struct(self.interpreter_address)?;
-
-        let mut traces =
-            get_stack_traces(&interp, &self.core, self.threadstate_address, Some(config))?;
-        let thread_names = thread_names_from_interpreter(&interp, &self.core, &self.version).ok();
+        let mut traces = get_stack_traces::<I, CoreDump>(
+            self.interpreter_address,
+            &self.core,
+            self.threadstate_address,
+            Some(config),
+        )?;
+        let thread_names = thread_names_from_interpreter::<I, CoreDump>(
+            self.interpreter_address,
+            &self.core,
+            &self.version,
+        )
+        .ok();
 
         for trace in &mut traces {
             if let Some(ref thread_names) = thread_names {
--- a/src/python_interpreters.rs
+++ b/src/python_interpreters.rs
@@ -24,8 +24,10 @@ pub trait InterpreterState {
     type TupleObject: TupleObject;
     const HAS_GIL_RUNTIME_STATE: bool = false;
 
-    fn head(&self) -> *mut Self::ThreadState;
-    fn modules(&self) -> *mut Self::Object;
+    /// Get a remote pointer to a pointer to PyThreadState.
+    fn threadstate_ptr_ptr(interpreter_address: usize) -> *const *const Self::ThreadState;
+    /// Get a remote pointer to a pointer to PyObject being the modules dict.
+    fn modules_ptr_ptr(interpreter_address: usize) -> *const *const Self::Object;
 }
 
 pub trait ThreadState {
@@ -115,11 +117,13 @@ macro_rules! PythonCommonImpl {
             type ListObject = $py::PyListObject;
             type TupleObject = $py::PyTupleObject;
 
-            fn head(&self) -> *mut Self::ThreadState {
-                self.tstate_head
+            fn threadstate_ptr_ptr(interpreter_address: usize) -> *const *const Self::ThreadState {
+                (interpreter_address + std::mem::offset_of!(Self, tstate_head))
+                    as *const *const Self::ThreadState
             }
-            fn modules(&self) -> *mut Self::Object {
-                self.modules
+            fn modules_ptr_ptr(interpreter_address: usize) -> *const *const Self::Object {
+                (interpreter_address + std::mem::offset_of!(Self, modules))
+                    as *const *const Self::Object
             }
         }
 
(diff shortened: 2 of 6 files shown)


benfred:
> I see failed builds due to an older rustc version in use 

I've recently updated the version of rust on the docker images and freebsd builds - hoping that re-running fixes

benfred (approved):
thanks for the fix! this looks good to me.

as a bit of an aside, I'm starting to think about leveraging the _Py_DebugOffsets struct that was introduced in python 3.12  (like https://github.com/python/cpython/blob/2a87af062b79d914ce0120f1f1763213c1ebe8c4/Include/internal/pycore_debug_offsets.h#L67-L96)  -  as a method to support things like freethreaded python builds and debug build pythons.  Your change here seems to be in line with the work needed for that - which is great

karolinepauls:
Thanks - `_Py_DebugOffsets` will be a definite improvement.


I've been curious about the performance difference so I wrote a small benchmark measuring running `runner.spy.get_stack_traces()` on a 50-calls deep single-threaded Python program.

Sorry for a screenshot, it's all I have left.
image

The disparity between 3.13 and 3.12 is removed.
- 3.13 takes 0.83% of the time the previous version took
- 3.12 takes 0.73%

The benchmark is here in case you want to consider merging it: https://github.com/benfred/py-spy/compare/master...karolinepauls:py-spy:bench?expand=1

PR #791, opened 2025-07-31
Add the Copy trait for python interpreter traits

benfred:
We are doing a bitwise copy from the target python process for various python structs - but weren't saying that these structs must implement Copy. Fix.


benfred:
fwiw - this is required by changes here https://github.com/benfred/remoteprocess/pull/109

PR #800, opened 2025-09-04
[Hardware][RISC-V] Add riscv64 support

langc23:
The purpose of this PR is to support RISC-V architectures.


langc23:
@benfred Hello, we hope the PR could merged into the mainline, but need your review and start workflow. Could you please take a time? Thanks.

langc23:
@benfred @yihong0618 @peytondmurray @zanieb  Hello, we hope the PR could merged into the mainline, but need your review and start workflow. Could you please take a time? Thanks.

langc23:
@benfred @yihong0618 @peytondmurray @zanieb Hello, the purpose of this PR is to support RISC-V architectures. We hope the PR could merged into the mainline, but need your review and start workflow. Could you please take a time? Thanks.

peytondmurray:
@langc23 Sorry, I'm not a maintainer, just an occasional contributor :shrug:

yihong0618:
@langc23 Sorry, me too...

yihong0618:
@langc23 just checked its MIT License if you indeed need it, you can fork it and publish to pypi yourself

maybe named py-spy-fork and waiting for the maintainer not so busy time

langc23:
> @langc23 just checked its MIT License if you indeed need it, you can fork it and publish to pypi yourself
> 
> maybe named py-spy-fork and waiting for the maintainer not so busy time

Thank you for your suggestions. But I also hope it could merged into the mainline.

src/python_bindings/v3_11_6.rs
```diff
@@ -0,0 +1,2135 @@
+/* automatically generated by rust-bindgen 0.71.1 */
```
benfred, line 1: Why do you need a separate v3.11.6 bindings here? We've managed to use the existing v3.11 bindings for x86_64/aarch64/armv7/i686 cpu's - and I'm curious why this won't work for risc64? 

Also does this mean that this risc support will only work with python 3.11?

src/python_bindings/mod.rs
```diff
@@ -1,6 +1,7 @@
 pub mod v2_7_15;
 pub mod v3_10_0;
 pub mod v3_11_0;
+pub mod v3_11_6;
```
benfred, line 4: How is this v3_11_6 module being used? We've implemented the traits for all the bindings - and just importing the module on its own won't do much

benfred (changes requested):
How is this tested? Can we verify that this works in CI against a risc64 cpu, and with more python versions than just v3.11.6 ?

src/python_bindings/mod.rs
langc23: This module is adapted for the RISC-V 64-bit architecture, e.g. openEuler 24.03.

src/python_bindings/v3_11_6.rs
langc23: Because the existing binding does not specify the RISC-V architecture. But we add [cfg(all(target_os = "linux", target_arch = "riscv64"))].

It is also possible to modify it to support x86 and other architectures, but since the Python version on RISC-V 64 is generally 3.11.6, I chose this version.

src/python_bindings/mod.rs
benfred: My point was that this module is dead code - you've generated the bindings, but aren't using them anywhere.  

If this actually was necessary you'd have to update places like https://github.com/benfred/py-spy/blob/194bf644dcae916587bf81daaefd25bfe0b737f1/src/python_interpreters.rs#L691  and https://github.com/benfred/py-spy/blob/194bf644dcae916587bf81daaefd25bfe0b737f1/src/python_spy.rs#L165-L169 like the actual v3_11_0 code does to use this file

This entire file needs deleted from the PR as its not used anywhere.

langc23: Yes, you are right. I have changed the commit, please review it again. Thanks!

PR #805, opened 2025-10-11
fix: print error twice

### LINKED ISSUE #804: if no pid the py-spy will print the error twice
```cli
➜  py-spy git:(master) sudo py-spy dump -p 555
Error: Failed to open process - check if it is running.
Reason: Operation not permitted (os error 1)
Reason: Operation not permitted (os error 1)
```

yihong0618:
close #804 

py-spy use anyhow for error and it already display as the method 

```rs
        for (i, suberror) in err.chain().enumerate() {
            if i > 0 {
                eprintln!("Reason: {suberror}");
            }
        }
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -510,11 +510,6 @@ fn main() {
         }
 
         eprintln!("Error: {err}");
-        for (i, suberror) in err.chain().enumerate() {
-            if i > 0 {
-                eprintln!("Reason: {suberror}");
-            }
-        }
         std::process::exit(1);
     }
 }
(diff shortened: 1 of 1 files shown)


benfred:
Thanks for the PR! Its a good callout that we are displaying the errors twice

I've updated recently to use the debug formatter from anyhow directly https://github.com/benfred/py-spy/pull/841, rather than try to output each suberror/context ourselves. This lets us also display a stacktrace if ` RUST_BACKTRACE=1` is set on the environment - but does mean that this PR is probably not needed anymore

PR #825, opened 2026-01-21
feat: Format function names with qualname for Python 3.11+

jake-kramer:
This change will update function names to have module paths in them which is useful for disambiguating functions when viewing a profile.

See https://docs.python.org/3/glossary.html#term-qualified-name


benfred:
Hi @jake-kramer - this PR looks good to me,  can you update to the latest main branch and I will try to start CI on this ? (I don't seem to be able to update your branch from here)

jake-kramer:
Thanks @benfred, I can get to this next week

jake-kramer:
@benfred Ready for CI

benfred (approved):
thanks for the PR!

PR #819, opened 2025-12-19
Add support for Python 3.14

czardoz:
# Summary of changes

Had to jump through some hoops (hand-held by Claude & Codex).

`generate_bindings.py` + `v3_14_0.rs`:
- Suppressed `unnecessary_transmutes` warnings because bindgen generates conservative transmute calls that Rust 1.91+ flags.

`python_interpreters.rs`:
- Added 3.14 `FrameObject` impl: Mask tag bits from `_PyStackRef.bits` because Python 3.14 uses tagged pointers instead of raw pointers for `f_executable` and local variables.
  - *Ref: [gh-117139: Set up tagged pointers in the evaluation stack](https://github.com/python/cpython/issues/117139)*
  - *Ref: [gh-124218: Reference count contention with nested functions](https://github.com/python/cpython/issues/124218)*
- Fixed line number calculation (line 361): Changed `>=` to `>` because algorithm needs to process the entry where `bytecode_address == lasti` to get the correct line for that offset (was returning previous range's line number, causing intermittent test failures in Python 3.11-3.14).

`python_data_access.rs`:
- Refactored dict iteration: Python 3.14 changed inline values layout (stored after `tp_basicsize` instead of after `PyObject`) and uses `PyDictKeyEntryHash` for `dk_kind==0` instead of legacy format.
  - *Ref: [gh-115789: Remove ma_version_tag from PyDictObject](https://github.com/python/cpython/issues/115789)*
  - *Ref: [gh-116032: Add PyDictKeyEntryHash](https://github.com/python/cpython/issues/116032)*

`python_threading.rs`:
- Added fallback to `sysdict` for thread name lookup because Python 3.14 can fail to import threading module into main `sys.modules` in some cases.

`stack_trace.rs`:
- Use `F::read_local_address()` trait method because Python 3.14 local variables are `_PyStackRef` (tagged pointers requiring masking).
  - *Ref: [gh-117139: Set up tagged pointers in the evaluation stack](https://github.com/python/cpython/issues/117139)*

`tests/subprocesses.py`:
- Force fork context because Python 3.14+ defaults to forkserver which creates extra intermediate process, breaking test expectations.
  - *Ref: [gh-84559: Change default multiprocessing start method to forkserver on Linux](https://github.com/python/cpython/issues/84559)*

`tests/integration_test.rs`:
- Added error message to `test_recursive` assertion for debugging line number issues.

PR #831, opened 2026-03-28
Add Python 3.14 support via _Py_DebugOffsets

charles-dyfis-net:
Instead of generating version-specific bindgen struct bindings for 3.14, read interpreter state using field offsets from CPython's `_Py_DebugOffsets` metadata table (embedded at the start of `_PyRuntime`). This table, present since 3.13, provides offsets for all core interpreter structs -- `PyInterpreterState`, `PyThreadState`, `_PyInterpreterFrame`, `PyCodeObject`, and common Python object types -- allowing a profiler to read them without compile-time knowledge of their layouts.

The offset-based approach handles 3.14's key structural changes:

  - `_PyStackRef` tagged pointers replacing `PyObject*` in frame fields (extraction: `bits & !1`; null: `bits == 1`; tagged int: `bits & 3 == 3`)
  - `FRAME_OWNED_BY_CSTACK` renumbered from 3 to 4
  - Free-threaded `PyObject` header enlargement (16 -> 32 bytes)
  - Free-threaded `PyASCIIObject.state.interned` changed from a 2-bit bitfield to a full `unsigned char` (for atomic access), shifting kind/compact/ascii into the next byte
  - Thread-local bytecode (TLBC) in free-threaded builds, where `instr_ptr` may point into a per-thread copy rather than `co_code_adaptive`
  - Dict entry format change (`dk_kind==0` now uses a 3-field `KeyWithHash` entry instead of the legacy `PyDictKeyEntry`)
  - Inline values for object attributes (`Thread._name` is stored inline rather than in a materialized `__dict__`)

Feature parity with the trait-based path:

  - Stack traces with filenames, function names, and line numbers
  - GIL detection (including free-threaded builds where GIL is disabled)
  - Thread names (via offset-based dict iteration and inline value reading)
  - Local variable inspection (`--dump-locals`), including dict, list, tuple, string, int, float, and numpy scalar formatting
  - Native stack merging, active/idle detection, short filenames

Python 3.13 also uses the offset-based path (validated via oracle comparison tests against the trait-based path on identical frozen process state).

Performance: frame and thread state structs are bulk-read in single `process_vm_readv` / `mach_vm_read` calls via `StructBuf`, matching the trait-based path's existing behavior of ~2 struct reads per frame (frame + code object), plus string/linetable reads that both paths require.

The existing trait-based code path is unchanged for Python <=3.12. The 3.13 bindgen types (`v3_13_0`) are reused in a few places where `_Py_DebugOffsets` doesn't yet provide sufficient information:

  - `PyDictKeysObject` layout (no `PyObject` header, stable across builds)
  - `ht_cached_keys` offset in `PyHeapTypeObject` (derived from the 3.13 bindgen offset, adjusted by the `PyObject` header size delta for free-threaded builds)
  - `tp_dictoffset` within `PyTypeObject` (discovered at runtime by scanning a module type for a known value)

These workarounds are documented and will be unnecessary if python/cpython#146462 is resolved (adding `tp_dictoffset`, `tp_basicsize`, and `ht_cached_keys` to `_Py_DebugOffsets`).

Known limitations:

  - 32-bit platforms: inline value reading (`ht_cached_keys`) is not supported; thread names will be unavailable for 3.14 on 32-bit.
  - The `_dictvalues` header size (4 bytes padded to pointer alignment) and `MANAGED_DICT_OFFSET` (-1 or -3 * ptr_size depending on `Py_GIL_DISABLED`) are derived from CPython internals not exposed via `_Py_DebugOffsets`.
  - `PyCompactUnicodeObject` size is derived as `asciiobject_size + 2 * ptr_size` (for the `utf8_length` and `utf8` fields); `_Py_DebugOffsets` provides `asciiobject_size` but not `PyCompactUnicodeObject` size directly.

### DIFF
--- a/src/debug_offsets.rs
+++ b/src/debug_offsets.rs
@@ -1,0 +1,652 @@
+use anyhow::Error;
+use remoteprocess::ProcessMemory;
+
+use crate::python_bindings::v3_13_0;
+use crate::version::Version;
+
+/// Version-agnostic struct holding all debug offsets py-spy needs.
+/// Populated from `_Py_DebugOffsets` read from the target process.
+#[derive(Debug, Clone, Default)]
+#[allow(dead_code)]
+pub struct DebugOffsets {
+    pub version: u64,
+    pub free_threaded: bool,
+
+    // Runtime state
+    pub runtime_interpreters_head: u64,
+
+    // Interpreter state
+    pub interp_size: u64,
+    pub interp_threads_head: u64,
+    pub interp_imports_modules: u64,
+    pub interp_gil_runtime_state: u64,
+    pub interp_gil_runtime_state_enabled: u64,
+    pub interp_gil_runtime_state_locked: u64,
+    pub interp_gil_runtime_state_holder: u64,
+
+    // Thread state
+    pub thread_size: u64,
+    pub thread_prev: u64,
+    pub thread_next: u64,
+    pub thread_interp: u64,
+    pub thread_current_frame: u64,
+    pub thread_id: u64,
+    pub thread_native_thread_id: u64,
+
+    // Interpreter frame
+    pub frame_size: u64,
+    pub frame_previous: u64,
+    pub frame_executable: u64,
+    pub frame_instr_ptr: u64,
+    pub frame_localsplus: u64,
+    pub frame_owner: u64,
+    pub frame_tlbc_index: u64,
+
+    // Code object
+    pub code_size: u64,
+    pub code_filename: u64,
+    pub code_name: u64,
+    pub code_qualname: u64,
+    pub code_linetable: u64,
+    pub code_firstlineno: u64,
+    pub code_argcount: u64,
+    pub code_localsplusnames: u64,
+    pub code_localspluskinds: u64,
+    pub code_co_code_adaptive: u64,
+    pub code_co_tlbc: u64,
+
+    // PyObject
+    pub pyobject_size: u64,
+    pub pyobject_ob_type: u64,
+
+    // Type object
+    pub type_object_tp_name: u64,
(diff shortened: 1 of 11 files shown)


charles-dyfis-net:
Adding Python 3.15 support should be straightforward after python/cpython#146462 lands. (Granted, Python 3.15 has its own high-performance sampling profiler -- but there are some features, such as native frame unwinding, that I don't know it to target, and having a single Rust library that can be used to build tools that monitor a wide range of Python interpreter releases has value all its own).

charles-dyfis-net:
BTW, I quite appreciate the prior work done in #819 by @czardoz. The value of going the route used here is in futureproofing and avoiding the need for tens of thousands of lines of generated code to be added for every new interpreter release.

Keeping this in draft for now pending further testing -- it's been developed and smoketested on aarch64 (in a tree also having e04be072df5021e3739f06c740a4070c445f57c5 applied); I'll want to spend some time with Intel platforms before calling it good.

santagada:
For what is worth I build this change and on windows with python 3.14 this never finds any stacks.

```
target\release\py-spy.exe record --pid 35860 -o test
py-spy> Sampling process 100 times a second. Press Control-C to exit.


py-spy> Stopped sampling because Control-C pressed
[2026-04-14T08:37:29.787074900Z ERROR inferno::flamegraph] No stack counts found
Error: Failed to write flamegraph: No stack counts found
```

charles-dyfis-net:
Interesting. Windows is not a platform I generally target -- will need to make an effort to get set up for development there.

charles-dyfis-net:
Since there's upstream support for 3.14, current plan is to wait for CPython to merge the extra requested offsets into the 3.15 tree, and target that only.

PR #842, opened 2026-04-23
Fix intermittent OSX CI errors

benfred:
We were seeing intermittent failures in OSX CI. This seems to be because the symbol address lookup was returning invalid addresses on OSX .

We check to make sure the interpreter address is valid after getting the address from symbols, and if it isn't we fallback to scanning the BSS section for the interpreter.  On OSX, this fallback wasn't always getting triggered - if the _PyRuntime address pointed to valid memory we would detect that the it was invalid and trigger the fallback - but if it pointed to an invalid address in the target process we just returned an error and failed to load. This caused intermittent failures on CI. This fix makes it so that we will run the BSS fallback even if the _PyRuntime address points to invalid memory.

Fixing the OSX symbol resolution so that we get valid addresses will happen in a separate PR (and is also blocking changes like https://github.com/benfred/py-spy/pull/835)

### DIFF
--- a/src/python_process_info.rs
+++ b/src/python_process_info.rs
@@ -373,86 +373,20 @@ where
 {
     // get the address of the main PyInterpreterState object from loaded symbols if we can
     // (this tends to be faster than scanning through the bss section)
-    match version {
-        Version {
-            major: 3,
-            minor: 13..=14,
-            ..
-        } => {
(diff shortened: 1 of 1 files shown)


benfred:
symbol fix is https://github.com/benfred/py-spy/pull/843

Issue #846, opened 2026-05-01
`dump --subprocesses` aborts on first non-Python child instead of skipping


ravwojdyla:
When `py-spy dump --pid <PID> --subprocesses` walks a process tree that contains a non-Python child (e.g. a Go binary), py-spy fails the entire command with `Error: Failed to find python version from target process` and exits non-zero - even though the parent and any earlier siblings were dumped successfully.

PR #847, opened 2026-05-01
Don't abort `dump --subprocesses` on non-Python children

ravwojdyla:
* https://github.com/benfred/py-spy/issues/846
* `dump --subprocesses` no longer aborts when a child isn't Python — bring it in line with `record`/`top`
* swap `PythonSpy::new` → `PythonSpy::retry_new(..., 5)` so a still-initializing child gets the same retry window the live samplers grant
* catch errors from the recursive `write_traces` call and `warn!` instead of `?`-propagating [^1]
* hoist the `parentpid != pid` skip above the separator `writeln!` so grandchildren no longer print a stray `----`
* regression test: parent + python child + non-python sibling (`sh -c 'sleep 3600'`); asserts `write_traces` returns `Ok`, parent and python child appear, non-python child does not

  [^1]: matches `Sampler::new_subprocess_sampler` in `src/sampler.rs` — invisible by default, visible with `RUST_LOG=warn`

### DIFF
--- a/src/dump.rs
+++ b/src/dump.rs
@@ -2,6 +2,7 @@ use std::io::Write;
 
 use anyhow::{Context, Error};
 use console::{style, Term};
+use log::warn;
 
 use crate::config::Config;
 use crate::python_spy::PythonSpy;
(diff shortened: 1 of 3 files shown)


ravwojdyla:
> [Build / Test Wheels (3.8.0, windows-latest) (pull_request)](https://github.com/benfred/py-spy/actions/runs/25219208888/job/73947605549?pr=847)Failing after 13s

That looks flaky to me ?

ravwojdyla:
👋 @benfred and @santagada please take a look 🙇

PR #848, opened 2026-05-04
 Gate CLI dependencies behind optional `cli` feature flag

korniltsev-grafanista:
py-spy is used as a library in pyroscope-rs where CLI dependencies
(clap, inferno, console, ctrlc, indicatif, env_logger) are unnecessary.
This adds a `cli` feature flag (enabled by default) that guards all
CLI-only dependencies, allowing library consumers to opt out with
`default-features = false`.

  Changes:
  - Make clap, clap_complete, console, ctrlc, indicatif, inferno, and
    env_logger optional dependencies gated by the `cli` feature
  - Add `[[bin]]` with `required-features = ["cli"]` so the binary is
    skipped when building without CLI
  - Gate clap-dependent code in config.rs (ArgEnum derive, from_commandline,
    from_args, CLI tests) behind cli feature
  - Add CI step to build library without cli feature
  - Gate coredump and dump modiles behind cli feature

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,6 +43,8 @@ jobs:
           python-version: 3.9
       - name: Build
         run: cargo build --release --verbose --examples
+      - name: Build library
+        run: cargo build --release --verbose --lib --no-default-features
       - uses: actions/setup-python@v6
         with:
           python-version: 3.9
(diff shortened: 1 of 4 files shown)


benfred (approved):
thanks for the PR! 

This does make me wonder if the dump/coredump code should even be exported from `src/lib.rs` at all - but I think we can leave as is for now

korniltsev-grafanista:
@benfred Thanks for prompt review 🙏 💘 
Any chance to publish a new version of the crate for downstream library users?

benfred:
There are a couple of other things I'd like to get into a new release - but I'm aiming for sometime in the next two weeks

PR #854, opened 2026-05-28
fix: use debug offsets for modules on 3.14

### LINKED ISSUE #852: Failed to get thread names on python v3.14.5
Thread names aren't being returned on python v3.14.5:  https://github.com/benfred/py-spy/pull/851

This seems to be because of `PyInterpreterState.imports.modules` being at a different location versus v3.14.4. We should handle this case, possibly by leveraging `Py_DebugOffsets.interpreter_state.imports_modules` to get the correct location

ilia-eiq:
Attempt to fix #852 using DebugOffsets, not tested yet

### DIFF
--- a/src/python_interpreters.rs
+++ b/src/python_interpreters.rs
@@ -23,12 +23,17 @@ pub trait InterpreterState: Copy {
     type StringObject: StringObject;
     type ListObject: ListObject;
     type TupleObject: TupleObject;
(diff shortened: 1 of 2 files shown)


ilia-eiq:
Nevermind, I misunderstood how pyruntime_addr and interpreter_address relate, so this doesn't work

neeehvns-tech (approved):

PR #855, opened 2026-06-02
fix: support 3.14.5 with generated bindings

ilia-eiq:
Brute force fix for #852 

Generated and fixed new bindings for 3.14.5, copied implementation for 3.14.0


jonashaag:
@charles-dyfis-net can we get this merged and released for 3.14? I can't use py-spy with 3.14 right now because it keeps attempting to read thread names over and over again, making it fall behind on sampling. Codex came up with this https://github.com/benfred/py-spy/pull/861 but I take it the solution in https://github.com/benfred/py-spy/pull/831 is more general and thus preferred.

